PR #28, opened 2019-07-01
initial secure memory implementation write-up

jpaskhay:
- we'll likely need to deal with some merge conflicts in this file


sushantmimani:
Will reassess if we need to split the contents of Internals.md further when we begin reviewing the algorithms pseudocode section

PR #35, opened 2019-07-08
Algorithm psuedocode

sushantmimani:


README.md
```diff
@@ -138,7 +138,8 @@ Optional<JSONObject> payload = appEncryptionJson.load(persistenceKey, dataPersis
 * [Key Management Service](docs/KeyManagementService.md)
 * [Metastore Persistence](docs/Metastore.md)
 * [Key Caching](docs/KeyCaching.md)
-* [Common APIs and Algorithm Internals](docs/Internals.md)
+* [Directory Structure and Common APIs](docs/DirectoryStructureAndCommonAPIs.md)
```
jpaskhay: given our refactoring of this content, i think it would be cleaner to simplify/generalize this to `Code Structure` or similar (ditto on filename and main header). also avoids coupling to the headers in case we add/edit more later

docs/Internals.md
```diff
@@ -1,150 +1,165 @@
...
-    <T> T withDecryptedKey(byte[] keyCipherText, Instant keyCreated, boolean revoked,
-                           BiFunction<CryptoKey, Instant, T> actionWithDecryptedKey);
-}
-```
-
-Detailed information about the  Key Management Service can be found [here](KeyManagementService.md) 
-
+# SDK Internals
+
+## Algorithm Internals
```
jpaskhay: i think your original `Envelope Encryption Algorithm` header works better here (it's more specific to the resulting content)

docs/Internals.md
```diff
@@ -1,150 +1,165 @@
...
+    If IK meta in message = current IK in metadata persistence
+        If SK EKR DOES NOT exist in metadata persistence 
+            THROW ERROR: no SK exists
+        Load current SK EKR from metadata persistence
+        Use MK in HSM to decrypt SK
+        If SK is expired
+            Queue SK for rotation
+            Queue IK for rotation
+        Else 
+            Create new IK from crypto library (i.e. openssl)
```
jpaskhay: nit: replace `i.e.` with `e.g.` (think we missed in the original changes prior to moving to public repo)

docs/Internals.md
```diff
@@ -1,150 +1,165 @@
...
+            Use MK in HSM to decrypt SK
+        If SK is expired
+            Queue SK for rotation
+            Queue IK for rotation
+            Exit #We'll be back once SK has rotated
+        Use SK to decrypt IK
+    If IK is expired
+        Queue IK for rotation
+        Exit  #We'll be back once IK has rotated
+    Create new DRK from crypto library (i.e. openssl)
```
jpaskhay: nit: replace `i.e.` with `e.g.` (think we missed in the original changes prior to moving to public repo)

docs/Internals.md
```diff
@@ -1,150 +1,165 @@
...
+        Load latest IK EKR from metadata persistence
+        If IK is found
+            If IK is not expired or (IK is expired and policy allows queued rotation)
+                If SK is not cached
+                    Load specific SK EKR from metadata persistence
+                    If SK EKR DOES NOT exist in metadata persistence
+                        Fall through to new IK creation
+                    If allowed by policy, add SK to protected memory cache
+                If SK is expired
+                    # NOTE: Possible inconsistency: when policy doesn't use inline rotation, consider proceeding without forced creation (same as IK handling)
```
jpaskhay: minor: 120 char line limit (may be another line or two in this file violating the limit, one of which i think i goofed)

docs/CodeStructure.md
```diff
@@ -0,0 +1,147 @@
...
+```
+
+Detailed information about the Metastore can be found [here](Metastore.md) 
+
+
+### Key Management Service
+
+```java
+  // Defines the root KMS
+  interface KeyManagementService {
```
nikoo28: unintentional extra space in the beginning?
Same at L:98 and L:122

nikoo28:
Also, we have 2 spaces at some places(java readme) and 4(c# readme and asherah's landing page) at the other. Should we make it consistent?

jpaskhay:
> Also, we have 2 spaces at some places(java readme) and 4(c# readme and asherah's landing page) at the other. Should we make it consistent?

We should stick with the language-specific styles we've chosen.

docs/CodeStructure.md
sushantmimani: fixed

nikoo28:
> > Also, we have 2 spaces at some places(java readme) and 4(c# readme and asherah's landing page) at the other. Should we make it consistent?
> 
> We should stick with the language-specific styles we've chosen.

Yep, that is what my concern was. Since Asherah's landing page has a JAVA example and it is indented with 4 spaces.

PR #40, opened 2019-07-10
added a couple implementation details to Internals

jpaskhay:
- added thread safety note regarding Secure Memory secret read
- added Key Cache implementation notes around TTL and duplicates

### DIFF
--- a/docs/Internals.md
+++ b/docs/Internals.md
@@ -309,8 +309,10 @@ to their [intended lifecycle](KeyCaching.md#cache-lifecycles)).
 
 Since the objects being cached are resources which need to be closed, there is additional complexity when dealing with
 duplicates in the cache. The approach taken in the Java reference implementation is to always return the key intended
-to be closed to the caller. For the case of a new key being added to the cache, we return a new "shared key"
-representation of the key whose close operation is a no-op (since the key passed in to the cache put/store call will
-now be used in the cache by other threads). For the case of a duplicate key being added to the cache (e.g. a race
-condition's second thread), we return the key passed in to the cache put/store call so it can be safely closed without
-affecting the existing underlying key in the cache and ensuring we don't leak the memory space of the key.
+to be closed to the caller:
+* For the case of a new key being added to the cache, we return a new "shared key" representation of the key whose
+close operation is a no-op (since the key passed in to the cache put/store call will now be used in the cache by other
+threads).
+* For the case of a duplicate key being added to the cache (e.g. a race condition's second thread), we return the key
+passed in to the cache put/store call so it can be safely closed without affecting the existing underlying key in the
+cache and ensuring we don't leak the memory space of the key.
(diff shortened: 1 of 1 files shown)


docs/Internals.md
```diff
@@ -254,6 +254,14 @@ withSecretBytes(function<byte[], type> functionWithSecret) {
 }
 ```
 
+##### Concurrent Access
+
+The `withSecretBytes` pseudocode above is not thread-safe code as written. A thread could disable the memory access as
+another thread attempts to read the secret, which would result in a SIGSEGV signal to the process. Some form of thread
+safety is needed to guard against this. In the Java reference implementation (as of this writing) this is implemented
```
nikoo28: do we need to list `as of this writing`? 
Would make the reader think if the documentation is actually up to date or not.

docs/Internals.md
```diff
@@ -286,3 +294,23 @@ library such as OpenSSL, BoringSSL, etc. The intent of this effort would be to s
...
+associated with interacting with a KMS/HSM provider (recall TTL is more likely to come into play with System Keys due
+to their [intended lifecycle](KeyCaching.md#cache-lifecycles)).
+
+### Duplicate Key Handling
+
+Since the objects being cached are resources which need to be closed, there is additional complexity when dealing with
+duplicates in the cache. The approach taken in the Java reference implementation is to always return the key intended
+to be closed to the caller. For the case of a new key being added to the cache, we return a new "shared key"
+representation of the key whose close operation is a no-op (since the key passed in to the cache put/store call will
+now be used in the cache by other threads). For the case of a duplicate key being added to the cache (e.g. a race
```
nikoo28: Can we make these 2 cases as 2 bullet points or just hyphens? I think it would look more clear.

jpaskhay: ya i thought about doing that at one point, and i think i simply forgot. will do so

docs/Internals.md
jpaskhay: the intent of that bit was so this doesn't end up being inaccurate in case the implementation did change and we forgot to update this. it was meant to just serve as an example of how it could be implemented. can remove if there is still strong objection

jpaskhay:
> Can we just say `Java implementation` instead of `Java reference implementation`. Example in L:311 and L:303

was using that verbiage to reinforce the fact that the Java implementation serves as the reference implementation in the context of this page. thought it would sound more intentional but may come across as repetitive, hmm

docs/Internals.md
sushantmimani: i suggest we keep it

nikoo28: I am still doubtful about it. The documentation should always be accurate regardless of any code change that we make.
If we are documenting something, it means it is there. 

Once we update the code, we can create a ticket/issue so that we don't forget updating. IMO `as of this writing` is not providing the reader any information.

jpaskhay: again, this wasn't intended to document the code, so think we're splitting hairs here. i'm going to reword this to something more generic to make everyone happy

PR #45, opened 2019-07-18
Deadlock bug fix

nikoo28:
The way INSERT query was written using (WHERE NOT EXISTS) caused deadlocks when multiple threads try to write the same value at the same time. 
Fixed by always returning a false if the query fails

### DIFF
--- a/languages/csharp/AppEncryption/AppEncryption/Persistence/AdoMetastorePersistenceImpl.cs
+++ b/languages/csharp/AppEncryption/AppEncryption/Persistence/AdoMetastorePersistenceImpl.cs
@@ -23,12 +23,10 @@ namespace GoDaddy.Asherah.AppEncryption.Persistence
 
         private const string LoadQuery = @"SELECT key_record from encryption_key where id = @id and created = @created";
 
-        // Store only if a value with the same key doesn't exist
-        // Using this approach since dotnet doesn't provide a specific integrity violation exception
+        // We try to store even if a value with the same key exists
+        // since dotnet doesn't provide a specific integrity violation exception
         private const string StoreQuery =
-            @"INSERT INTO encryption_key (id, created, key_record) 
-              SELECT @id, @created, @key_record WHERE NOT EXISTS 
-              (SELECT id from encryption_key where id = @id and created = @created )";
+            @"INSERT INTO encryption_key (id, created, key_record) SELECT @id, @created, @key_record";
 
         private const string LoadLatestQuery =
             @"SELECT key_record from encryption_key where id = @id order by created DESC limit 1";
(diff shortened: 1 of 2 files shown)


nikoo28:
PR

https://github.com/godaddy/asherah/pull/45

languages/csharp/AppEncryption/AppEncryption/Persistence/AdoMetastorePersistenceImpl.cs
```diff
@@ -23,12 +23,10 @@ public class AdoMetastorePersistenceImpl : IMetastorePersistence<JObject>
 
         private const string LoadQuery = @"SELECT key_record from encryption_key where id = @id and created = @created";
 
-        // Store only if a value with the same key doesn't exist
-        // Using this approach since dotnet doesn't provide a specific integrity violation exception
+        // We try to store even if a value with the same key exists
```
jpaskhay: we can probably omit the comment entirely since we're no longer doing anything special with the insert statement

languages/csharp/AppEncryption/AppEncryption/Persistence/AdoMetastorePersistenceImpl.cs
```diff
@@ -23,12 +23,10 @@ public class AdoMetastorePersistenceImpl : IMetastorePersistence<JObject>
...
+        // since dotnet doesn't provide a specific integrity violation exception
         private const string StoreQuery =
-            @"INSERT INTO encryption_key (id, created, key_record) 
-              SELECT @id, @created, @key_record WHERE NOT EXISTS 
-              (SELECT id from encryption_key where id = @id and created = @created )";
+            @"INSERT INTO encryption_key (id, created, key_record) SELECT @id, @created, @key_record";
```
jpaskhay: since we're just doing a plain insert statement now, we should use the standard `insert into (column1, ...) values (@param1, ...)` style

(the insert select style is typically only used for bulk inserts or where not exists)

languages/csharp/AppEncryption/AppEncryption/Persistence/AdoMetastorePersistenceImpl.cs
```diff
@@ -123,16 +121,16 @@ public bool Store(string keyId, DateTimeOffset created, JObject value)
...
                             return result == 1;
                         }
                     }
                 }
                 catch (DbException dbe)
                 {
                     Logger.LogError(dbe, "Metastore error during store");
-                    throw new AppEncryptionException("Metastore error:", dbe);
+
+                    // Even if the query failed due to a duplicate record return false
```
jpaskhay: may be worth expanding/clarifying this comment slightly. basically conveying that ado doesn't have a specific integrity violation error code/exception and we decided to always return false even for systemic issues for simplicity (system errors in select queries will blow up anyway)

PR #91, opened 2019-07-25
Add CI for C# repos

nikoo28:


.circleci/config.yml
```diff
@@ -124,3 +132,114 @@ jobs:
       - codecov/upload:
           file: languages/java/app-encryption/target/site/jacoco/jacoco.xml
           flags: languages_java_app_encryption
+  languages-csharp-logging:
+    docker:
+      - image: mcr.microsoft.com/dotnet/core/sdk:2.1
```
jpaskhay: why are we not using 2.2? we were using it when we were building internally

.circleci/config.yml
```diff
@@ -124,3 +132,114 @@ jobs:
...
+    steps:
+      - checkout
+      - run:
+          name: Generate cache key
+          command: |
+            cd languages/csharp/
+            ./checksum.sh /tmp/checksum.txt ./Logging
+      - restore_cache:
+          keys:
+            # when pom changes, use increasingly general patterns to restore cache.
```
jpaskhay: nit: update pom reference (or generalize)

.circleci/config.yml
```diff
@@ -124,3 +132,114 @@ jobs:
...
+      - run:
+          name: Tests
+          command: |
+            cd languages/csharp/Logging
+            ./scripts/test.sh
+      - save_cache:
+          paths:
+            - ~/.nuget
+          key: v1-nuget-{{ .Branch }}-{{ checksum "/tmp/checksum.txt" }}
+      - store_test_results:
```
jpaskhay: doesn't seem like this is working properly. according to https://circleci.com/docs/2.0/collect-test-data/ it seems like they need junit or cucumber format files. although not as critical as code coverage, it would still be nice if we can get the test result summary to show up in the circleci builds

nikoo28:
- Used 2.2 image
- Made general comment
- Used https://github.com/gfoidl/trx2junit to store test results for `dotnet`

nikoo28:
Should be merged before PR https://github.com/godaddy/asherah/pull/92

.circleci/config.yml
```diff
@@ -83,12 +98,12 @@ jobs:
           command: |
             cd languages/java/secure-memory
             ./scripts/test.sh
+      - store_test_results:
```
jpaskhay: why did the order of this step change? the intent was to group test-related uploading logically at the end

.circleci/config.yml
```diff
@@ -115,12 +135,171 @@ jobs:
...
+          name: Tests
+          command: |
+            cd languages/csharp/Logging
+            ./scripts/test.sh
+      - run:
+          name: Convert test results
+          command: |
+            cd languages/csharp/Logging
+            dotnet tool install -g trx2junit
+            export PATH="$PATH:/root/.dotnet/tools"
```
jpaskhay: you can just add `--tool-path /usr/bin` to the above command

.circleci/config.yml
```diff
@@ -115,12 +135,171 @@ jobs:
...
+          name: Tests
+          command: |
+            cd languages/csharp/SecureMemory
+            ./scripts/test.sh
+      - run:
+          name: Convert test results
+          command: |
+            cd languages/csharp/SecureMemory
+            dotnet tool install -g trx2junit
+            export PATH="$PATH:/root/.dotnet/tools"
```
jpaskhay: ditto

.circleci/config.yml
```diff
@@ -115,12 +135,171 @@ jobs:
...
+          name: Tests
+          command: |
+            cd languages/csharp/AppEncryption
+            ./scripts/test.sh
+      - run:
+          name: Convert test results
+          command: |
+            cd languages/csharp/AppEncryption
+            dotnet tool install -g trx2junit
+            export PATH="$PATH:/root/.dotnet/tools"
```
jpaskhay: ditto

.circleci/config.yml
```diff
@@ -115,12 +135,171 @@ jobs:
           command: |
             cd languages/java/app-encryption
             ./scripts/test.sh
+      - store_test_results:
```
jpaskhay: ditto

.circleci/config.yml
```diff
@@ -115,12 +135,171 @@ jobs:
...
+      - run:
+          name: Convert test results
+          command: |
+            cd languages/csharp/Logging
+            dotnet tool install -g trx2junit
+            export PATH="$PATH:/root/.dotnet/tools"
+            trx2junit Logging.Tests/TestResults/*.trx
+      - store_test_results:
+          path: languages/csharp/Logging/Logging.Tests/TestResults
+      - save_cache:
```
jpaskhay: i think this should be moved up to run sooner (once we know all possible dependencies have been pulled down). the Tests (test.sh) step should theoretically be the last possible place they could be pulled down.

this also has an added benefit of the convert test results, store_test_results and codecov steps all being logically grouped together

.circleci/config.yml
```diff
@@ -115,12 +135,171 @@ jobs:
...
+      - run:
+          name: Convert test results
+          command: |
+            cd languages/csharp/SecureMemory
+            dotnet tool install -g trx2junit
+            export PATH="$PATH:/root/.dotnet/tools"
+            trx2junit SecureMemory.Tests/TestResults/*.trx
+      - store_test_results:
+          path: languages/csharp/SecureMemory/SecureMemory.Tests/TestResults
+      - save_cache:
```
jpaskhay: ditto

.circleci/config.yml
```diff
@@ -115,12 +135,171 @@ jobs:
...
+      - run:
+          name: Convert test results
+          command: |
+            cd languages/csharp/AppEncryption
+            dotnet tool install -g trx2junit
+            export PATH="$PATH:/root/.dotnet/tools"
+            trx2junit AppEncryption.Tests/TestResults/*.trx
+      - store_test_results:
+          path: languages/csharp/AppEncryption/AppEncryption.Tests/TestResults
+      - save_cache:
```
jpaskhay: ditto

.circleci/config.yml
```diff
@@ -65,11 +75,16 @@ jobs:
       - image: circleci/openjdk:8-jdk
     steps:
       - checkout
+      - run:
+          name: Generate cache key
+          command: |
+            cd languages/java/app-encryption
+            ../../../scripts/checksum.sh /tmp/checksum.txt pom.xml
```
jpaskhay: 1) use `$CIRCLE_WORKING_DIRECTORY` instead of the relative path
2) nit: now that i think of it, wonder if `.circleci` directory may be even more appropriate than `scripts` (since it is purely for circleci at this point)?

.circleci/config.yml
nikoo28: I changed it to group all the test related stuff and container file system stuff together.
And once everything is done, upload results is the last step.

nikoo28:
Need to finalize on the the order of steps. I would suggest

- Checkout
- Generate cache
- restore cache
- build
- save cache
- run tests
- convert results
- store results
- upload results

jpaskhay:
save cache needs to occur after run tests as that step can potentially pull down more dependencies (e.g. java)

nikoo28:
- [ ] Corrected build steps order
- [ ] Added TODO because env variable does not expand https://discuss.circleci.com/t/circle-working-directory-doesnt-expand/17007/5
- [ ] Keeping the script in root folder
- [ ] Still exporting `PATH`

.circleci/config.yml
jpaskhay: discussed offline: we'll come back to this one, hence the TODO/note

.circleci/config.yml
```diff
@@ -65,11 +75,18 @@ jobs:
       - image: circleci/openjdk:8-jdk
     steps:
       - checkout
+      - run:
+          name: Generate cache key
+          # TODO : Need to use CIRCLE_WORKING_DIRECTORY
+          # Refer to https://discuss.circleci.com/t/circle-working-directory-doesnt-expand/17007/5
+          command: |
+            cd languages/java/app-encryption
```
jpaskhay: `secure-memory`

.circleci/config.yml
jpaskhay: discussed offline. we'll leave in place as the suggestion isn't compatible w/ macos, and we're trying to leave steps macos-compatible in case we reuse steps later in macos executors

PR #95, opened 2019-07-30
POM file updates for CD

sushantmimani:
- Update POM files for artifact deployment
- Add maven settings file GPG signing and artifact deployment


languages/java/app-encryption/pom.xml
```diff
@@ -7,6 +7,34 @@
...
   <version>1.0.0-SNAPSHOT</version>
+  <name>Asherah</name>
+  <description>
+      An application-layer encryption SDK that provides advanced encryption features and in depth defense against
+      compromise.
+  </description>
+  <url>https://github.com/godaddy/asherah</url>
+
+  <licenses>
+      <license>
```
jpaskhay: match the indent level in this file (for all changes in this file).

side note: we should probably do a sweep in other POMs and the new `settings.xml` to update/normalize to same indent level (probably better in separate PR). i know AE and SM have always been different (we kept forgetting to fix). pretty sure the maven XML file standard is 2 spaces (also preferred for not nesting too deeply IMO)

languages/java/settings.xml
```diff
@@ -0,0 +1,24 @@
+<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0
+                      https://maven.apache.org/xsd/settings-1.0.0.xsd">
+    <servers>
```
jpaskhay: consider using 2 space indentation

PR #97, opened 2019-08-01
Java cicd cont

sushantmimani:
- Install AE and SM projects locally
- Disable DynamoDb container tests


.circleci/config.yml
```diff
@@ -67,6 +69,15 @@ jobs:
           command: |
             cd languages/java/secure-memory
             ./scripts/test.sh
+      - run:
```
nikoo28: The save cache step should be moved up. Ideally, save the cache as soon as we run the tests. So that all the dependencies are pulled in.

.circleci/config.yml
```diff
@@ -104,6 +117,15 @@ jobs:
           command: |
             cd languages/java/app-encryption
             ./scripts/test.sh
+      - run:
```
nikoo28: same over here

.circleci/config.yml
```diff
@@ -42,6 +42,8 @@ jobs:
       - image: circleci/openjdk:8-jdk
     steps:
       - checkout
+      - attach_workspace:
+          at: /home/circleci
```
jpaskhay: can we not use `$HOME` instead?

.circleci/config.yml
```diff
@@ -67,6 +69,15 @@ jobs:
           command: |
             cd languages/java/secure-memory
             ./scripts/test.sh
+      - run:
+          name: Install locally
+          command: |
+            cd languages/java/secure-memory
+            ./scripts/install_local.sh
+      - persist_to_workspace:
+          root: /home/circleci
```
jpaskhay: ditto

tests/java/test-app/pom.xml
```diff
@@ -133,79 +128,79 @@
...
-          <contextDirectory>${project.build.directory}</contextDirectory>
-          <dockerConfigFile>${docker.config.file}</dockerConfigFile>
-          <verbose>true</verbose>
-          <skipDockerInfo>true</skipDockerInfo>
-          <buildArgs>
-            <JAR_FILE>${project.build.finalName}.jar</JAR_FILE>
-          </buildArgs>
-        </configuration>
-      </plugin>
+<!--      <plugin>-->
```
jpaskhay: how about putting all this in a profile instead? that way we can still build these locally for running the testapp

tests/java/test-app/pom.xml
```diff
@@ -19,16 +19,11 @@
     <com.zaxxer.hikaricp.version>2.6.1</com.zaxxer.hikaricp.version>
     <mockito.core.version>2.22.0</mockito.core.version>
     <elasticsearch.version>6.3.2</elasticsearch.version>
-    <docker.image.src.root>images</docker.image.src.root>
-    <docker.config.file>${env.DOCKER_CONFIG}/config.json</docker.config.file>
-    <!-- <docker.registry></docker.registry> -->
+<!--    <docker.image.src.root>images</docker.image.src.root>-->
```
jpaskhay: see other comment

.circleci/config.yml
sushantmimani: $HOME doesn't work with persist_to_workspace

jpaskhay: lame. fair enough

tests/java/test-app/pom.xml
```diff
@@ -286,4 +208,92 @@
       <version>${com.zaxxer.hikaricp.version}</version>
     </dependency>
   </dependencies>
+
+  <profiles>
+    <profile>
+      <id>build-local</id>
```
jpaskhay: nit: consider `docker` instead of `local` (since it's all docker related hooks)

tests/java/test-app/pom.xml
```diff
@@ -286,4 +208,92 @@
...
+      <activation>
+        <property>
+          <name>local</name>
```
jpaskhay: nit: ditto

PR #98, opened 2019-08-01
Csharp CI builds

nikoo28:
- add configuration for building all C# projects
- add local nuget configuration
- temporarily reduce code coverage
- disable container based unit tests
- add scripts to install nuget packages locally
- required changes in deploy scripts and cs project files.


languages/csharp/AppEncryption/scripts/install_local.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+set -e
+
+FULL_VERSION=$(xmllint --xpath "//Project/PropertyGroup/Version/text()" Directory.Build.props)
+TIME_STAMP=$(date "+%Y%m%d")
```
jpaskhay: we should have more than just the date. be good to match the resolution the maven snapshot builds go down to (seconds)

.circleci/config.yml
```diff
@@ -172,6 +172,12 @@ jobs:
       - image: mcr.microsoft.com/dotnet/core/sdk:2.2
     steps:
       - checkout
+      - run:
+          name: Install dependencies
```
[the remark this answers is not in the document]
nikoo28: How does `Install development tools` sound?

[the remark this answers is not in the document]
jpaskhay: i'd say leave out the `development` -- could be tools that aren't necessarily development tools per se

languages/csharp/AppEncryption/scripts/install_local.sh
nikoo28: Can we just use the epoch time?

jpaskhay: i think that's less intuitive when looking at it (especially if we do ever decide to deploy these down the road). should just be a matter of expanding the format you're specifying

.circleci/NuGet.Config
```diff
@@ -0,0 +1,8 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
nikoo28: keeping build configuration files in root level `build` directory

.circleci/config.yml
```diff
@@ -291,19 +332,33 @@ jobs:
...
+          root: /root
+          paths:
+            - .nuget
   samples-csharp-reference-app:
     docker:
       - image: mcr.microsoft.com/dotnet/core/sdk:2.2
     steps:
       - checkout
+      - run:
+          name: Install tools
```
jpaskhay: minor (just now noticed): don't think this is needed for the sample app

languages/csharp/AppEncryption/scripts/install_local.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+set -e
+
+FULL_VERSION=$(xmllint --xpath "//Project/PropertyGroup/Version/text()" Directory.Build.props)
+TIME_STAMP=$(date "+%Y%m%d%S")
```
jpaskhay: looks like we added the seconds but not the hours nor minutes? :)

.circleci/config.yml
```diff
@@ -172,6 +172,17 @@ jobs:
...
+          command: |
+            apt-get update
+            apt-get install -y libxml2-utils
+            dotnet tool install -g trx2junit
+      - run:
+          name: Add local nuget configuration
```
jpaskhay: concerned about the fragility here, as this is the only job that performs this step. whenever we get rid of the Logging project, someone will have to remember/know that this is the only job that performs this step and it can get lost easily. may be worth putting in every C# job to be more resilient

.circleci/config.yml
```diff
@@ -199,6 +205,20 @@ jobs:
...
+          command: |
+            mkdir -p /root/.nuget/NuGet/
+            cp .circleci/NuGet.Config /root/.nuget/NuGet/
+      - run:
+          name: Install locally
+          command: |
+            cd languages/csharp/Logging
+            ./scripts/install_local.sh
+      - persist_to_workspace:
+          root: /root
```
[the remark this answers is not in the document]
jpaskhay: looks like we can't and doing something like `../` would be as bad since it would make another assumption of working dir. boo-urns. marking resolved

.circleci/config.yml
nikoo28: Should we also remove the deploy script then from the ReferenceApp?

jpaskhay: either way. doesn't matter as much. just trying to trim down the circleci config length

PR #99, opened 2019-08-02
Circle deploy flow

jpaskhay:
NOTE: Merge AFTER #97

- verified flows locally and with circleci builds
- note we have chosen not to deploy master (alpha) builds for C# for now
- will likely come back and create aliases for this


jpaskhay:
> We can either create templates at a later stage, or create a `deploy_if_needed.sh` kind of script to avoid duplicates and chances of errors.

👍 Ya, as we discussed earlier this week, probably better to get all the open PRs merged in first to have holistic view of config before refactoring.

PR #106, opened 2019-08-06
Add support for running Ado and Dynamo tests in a local docker container

sushantmimani:


nikoo28:
Commit summary

- Use local dynamodb and mysql images in `CircleCI`
- Added circleci environment variable `CONTAINER_TYPE` to determine what to pick
- local builds would still use test cotainers.

jpaskhay:
well, `Convert.ToBoolean` may not be as flexible as i thought, specifically, so may need another method. but still think a similar approach simplifies it

PR #111, opened 2019-08-30
naming normalization final-ish sweep 

jpaskhay:
- renamed remaining instances of `AppEncryptionSessionFactory` and `AppEncryption` (as an entity)
- updated places where we were still using "metastore persistence" (left one or two places where it was still appropriate)
- tried normalizing most of the instance of `[M|m]etaStore` -- didn't update the places where it's currently tied to testapp config files though (avoid updating our internal helm charts)
- minor typo cleanup along the way


jpaskhay:
proceeding with merge. if we find any other refactoring needed, we can handle as needed

PR #116, opened 2019-11-01
Fixed the flow in parameterised tests

nikoo28:
- Verify that some method calls actually never happen
- Fixed the case of loading latest SK in encrypt metastore interactions


jpaskhay:
chatted offline about the newlines. forgot we had intentionally grouped those blocks together. will mark those as resolved

PR #117, opened 2019-12-12
Java session caching functionality

jpaskhay:
- uses Caffeine for LRU/LFU functionality
  - using time and size based eviction policies
  - since the sessions are stateful, we're making use of Caffeine's callback functionality to verify sessions are safe to remove (thread safe as well)
    - implemented using compute-based lookups and a usage counter
    - if a session is determined to still be in use, we flag it as having infinite expiry for the time-based eviction and having no weight to contribute to the size-based eviction
  - added fair amount of comments around the implementation especially with the callback flows and some of the non-obvious changes
  - the actual cached object is `EnvelopeEncryptionJsonImpl` due to generics/simplicity, but we're effectively caching the session objects (and thereby the IKs)
- added 3 new options to CryptoPolicy
  - `canCacheSessions`: defaults to false to maintain backwards compatibility
  - `sessionCacheMaxSize`
  - `sessionCacheExpireMillis`
  - added javadocs to BasicExpiringCryptoPolicy builder to highlight the current defaults
- included unit tests for as many of the different/useful eviction and multi threaded scenarios i could think of
- removed `SecureCryptoKeyMapFactory` (wasn't really needed, especially with these changes)
- some minor refactoring cleanup in related bits of these changes


ben-manes:
This looks like a really good usage of the pin / unpin ability. Since expiration occurs during cache activity, for sessions you may prefer more promptness if you expect idleness. In that case, the `scheduler(Scheduler.systemScheduler())` option could be beneficial. This will schedule a `cache.cleanUp()` based on the next expiration event on a background thread (provided in JDK9 if using the system scheduler).

languages/java/app-encryption/src/test/java/com/godaddy/asherah/appencryption/SessionFactoryTest.java
```diff
@@ -42,32 +49,677 @@
...
+        byte[] drr = session.encrypt(payload);
+
+        // Sleep to trigger the expiration
+        // TODO Consider refactoring the Caffeine usage to allow injecting a ticker if sleeping becomes unreliable
+        try {
+          Thread.sleep(sessionCacheExpireMillis * 3);
+        } catch (Exception e) {
+          fail(e.getMessage());
+        }
+
```
nikoo28: can verify here that even after timeout, we have 1 entry in the cache

languages/java/app-encryption/src/test/java/com/godaddy/asherah/appencryption/SessionFactoryTest.java
```diff
@@ -42,32 +49,677 @@
...
+      try (Session<byte[], byte[]> session = factory.getSessionBytes(testPartitionId)) {
+        byte[] decryptedPayload = session.decrypt(drr);
+
+        assertArrayEquals(payload, decryptedPayload);
+        // verify no metastore interactions in the decrypt flow (since IKs cached via session caching)
+        verify(metastoreSpy, never()).load(any(), any());
+      } catch (Exception e) {
+        fail(e.getMessage());
+      }
```
nikoo28: Can verify before closing the outer session that nothing is evicted from cache since the session is still in use.

languages/java/app-encryption/src/test/java/com/godaddy/asherah/appencryption/SessionFactoryTest.java
```diff
@@ -42,32 +49,677 @@
...
+
+      try {
+        pool.shutdown();
+        pool.awaitTermination(60, TimeUnit.SECONDS);
+      } catch (InterruptedException e) {
+        e.printStackTrace();
+      }
+
+      assertEquals(numTasks, tasksCompleted.sum());
+      assertEquals(1, factory.getSessionCache().estimatedSize());
```
nikoo28: Instead of verifying this, we can verify after L:557 that number of entries in cache is less than the maxCacheSize defined.

languages/java/app-encryption/src/test/java/com/godaddy/asherah/appencryption/SessionFactoryTest.java
```diff
@@ -42,32 +49,677 @@
...
```
nikoo28: Instead of verifying this, we can verify after L:604 that number of entries in cache is less than the maxCacheSize defined.

languages/java/app-encryption/src/main/java/com/godaddy/asherah/appencryption/SessionFactory.java
```diff
@@ -204,4 +333,50 @@ public SessionFactory build() {
...
+    public void close() {
+      // Instead of closing the session, we call the release function so it can atomically update the usage counter
+      releaseShared(key);
+    }
+
+    EnvelopeEncryptionJsonImpl getEnvelopeEncryptionJsonImpl() {
+      return envelopeEncryptionJsonImpl;
+    }
+
+    void incrementUsageTracker() {
```
cstockton: The bare external reference counting here seems like a potential place for memory leaks to occur. Maybe something in lang/ref/*Reference may be used instead?

languages/java/app-encryption/src/test/java/com/godaddy/asherah/appencryption/SessionFactoryTest.java
jpaskhay: that's what the metastore interaction verify is effectively testing, but i can add the extra check

languages/java/app-encryption/src/test/java/com/godaddy/asherah/appencryption/SessionFactoryTest.java
jpaskhay: similar comment as above. can add the extra check

languages/java/app-encryption/src/test/java/com/godaddy/asherah/appencryption/SessionFactoryTest.java
jpaskhay: this isn't clear to me. that shouldn't be possible. the `maxCacheSize` is 1 in this test, and the cache will already be at that size once you're in that try block that gets a `Session`

languages/java/app-encryption/src/test/java/com/godaddy/asherah/appencryption/SessionFactoryTest.java
jpaskhay: similar comment here on not being possible in this test case. had to remove this assertion anyway since we couldn't rely on the `estimatedSize` call (covered in comment in the test case)

jpaskhay:
> This looks like a really good usage of the pin / unpin ability. Since expiration occurs during cache activity, for sessions you may prefer more promptness if you expect idleness. In that case, the `scheduler(Scheduler.systemScheduler())` option could be beneficial. This will schedule a `cache.cleanUp()` based on the next expiration event on a background thread (provided in JDK9 if using the system scheduler).

Thanks! Had to piece together some bits from old issues in your repo, but it ended up working out. Good callout on the `Scheduler` usage. The code changes are minimal for adding that, but I just need to spend some time adding unit tests to ensure it doesn't lead to any unforeseen issues. I'll try adding to this PR but if it gets too hairy I'll create a new Issue for it instead.

languages/java/app-encryption/src/main/java/com/godaddy/asherah/appencryption/SessionFactory.java
jpaskhay: Are you referring to this specific part or the caching in general? Been a while since I was working deeply on this, but IIRC when I looked into the soft and weak reference options for the cache, neither were going to feasible for our usage, unfortunately. In general, it is important for users to close the `Session` and `SessionFactory` instances, as indicated in the documentation around usage. Using the cache will make it even more important to close resources properly as it will prevent the finalizer from ever collecting the `Session` objects (finalizer isn't guaranteed to run in the first place).

We can always hop on a Zoom call to dive deeper, if that's easier.

PR #127, opened 2020-03-13
Documentation Update

sushantmimani:


nikoo28:
> Looks good. What do you think about adding the godoc badge to the top of the readme?
> 
> [GoDoc](https://godoc.org/github.com/godaddy/asherah/go/securememory)

`+1` on that. Can we also just keep the badge and skip the text link?

PR #131, opened 2020-03-24
Add cross-language tests for Go

sushantmimani:
- Add cross-language tests for CI
- Update scripts and CICD pipeline


tests/cross-language/go/encrypt_test.go
```diff
@@ -0,0 +1,123 @@
...
+		panic(e)
+	}
+
+	encryptedData, _ := json.Marshal(dataRow)
+	encryptedPayloadString = base64.StdEncoding.EncodeToString(encryptedData)
+
+	return nil
+}
+
+func ishouldgetencryptedData() error {
```
nikoo28: same over here

tests/cross-language/go/decrypt_test.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	encryptedPayload = string(data)
+
+	return nil
+}
+
+func idecrypttheencryptedData() error {
+	crypto := aead.NewAES256GCM()
+	manager, _ := kms.NewStatic(KeyManagementStaticMasterKey, crypto)
```
nikoo28: missed error check.

tests/cross-language/go/decrypt_test.go
```diff
@@ -0,0 +1,88 @@
...
+	defer sess.Close()
+
+	var dataRow appencryption.DataRowRecord
+
+	dataRowBytes, err := base64.StdEncoding.DecodeString(encryptedPayload)
+	if err != nil {
+		panic(err)
+	}
+
+	err2 := json.Unmarshal(dataRowBytes, &dataRow)
```
nikoo28: can just use `err`

tests/cross-language/go/const.go
```diff
@@ -0,0 +1,25 @@
...
+
+const (
+	KeyManagementStaticMasterKey = "mysupersecretstaticmasterkey!!!!"
+
+	DefaultServiceID   = "service"
+	DefaultProductID   = "product"
+	DefaultPartitionID = "partition"
+
+	FileDirectory = "/tmp/"
+	FileName      = "go_encrypted"
```
aka-bo: As we're only dealing with a single package here there's no need to export any of these

tests/cross-language/go/const.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+var (
+	MysqlDatbaseName = os.Getenv("TEST_DB_NAME")
+	MysqlUsername    = os.Getenv("TEST_DB_USER")
+	MysqlPassword    = os.Getenv("TEST_DB_PASSWORD")
+
+	ConnectionString = fmt.Sprintf("%s:%s@tcp(localhost:3306)/%s", MysqlUsername, MysqlPassword, MysqlDatbaseName)
```
aka-bo: Better to use a function here (also unexported)...

```go
func connectionString() string {
    return fmt.Sprintf(
        "%s:%s@tcp(localhost:3306)/%s",
        os.Getenv("TEST_DB_NAME"),
        os.Getenv("TEST_DB_USER"),
        os.Getenv("TEST_DB_PASSWORD"))
}
```

tests/cross-language/go/decrypt_test.go
```diff
@@ -0,0 +1,88 @@
...
+		Product: DefaultProductID,
+		Policy:  policy,
+	}
+
+	factory := appencryption.NewSessionFactory(config, metastore, manager, crypto)
+	defer factory.Close()
+
+	sess, err := factory.GetSession(DefaultPartitionID)
+	if err != nil {
+		panic(err)
```
aka-bo: Is there a benefit to using `panic` here (and below) instead of `return err`? Or `godog` convention?

tests/cross-language/go/encrypt_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+func encryptedDatashouldnotbeequaltodata() error {
+	if payloadString == encryptedPayloadString {
+		return errors.New("Encryption failure")
+	}
+
+	return nil
+}
+
+// nolint: deadcode
```
aka-bo: Is this needed? Seems odd that an exported function would be considered unused.

tests/cross-language/go/.gitignore
```diff
@@ -0,0 +1,3 @@
+.idea/
+bin/
+*.out
```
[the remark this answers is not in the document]
sushantmimani: Added .DS_Store. Don't think the other are needed since we don't generate code coverage/test results. Suggest we add them if we make such changes.

tests/cross-language/go/encrypt_test.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+var (
+	payloadString          string
+	encryptedPayloadString string
+	connection             *sql.DB
+)
+
+func iHave(payload string) error {
+	payloadString = payload
```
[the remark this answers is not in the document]
sushantmimani: payload is part of the feature and therefore will never be nil. Don't see a benefit of having that check here

tests/cross-language/go/decrypt_test.go
sushantmimani: Shadows the previous err, if used.

tests/cross-language/go/decrypt_test.go
sushantmimani: returning error now. A lot of this was copy-paste from the referenceapp. Should we remove the `panic` from there as well?

tests/cross-language/go/decrypt_test.go
nikoo28: the previous error has already been handled. can clarify, how it is being shadowed.

tests/cross-language/go/encrypt_test.go
sushantmimani: Was able to run everything successfully without needing this

tests/cross-language/java/scripts/build.sh
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+mvn -U compile
```
nikoo28: we shouldn't do `-U`. The cross language tests should use the `appencryption` jar from the restored cache.
`-U` forces an update on all the snapshots

tests/cross-language/go/decrypt_test.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	err2 := json.Unmarshal(dataRowBytes, &dataRow)
+	if err2 != nil {
+		return err2
+	}
+
+	data, err := sess.Decrypt(dataRow)
+
+	if err != nil {
```
nikoo28: this statement should be coupled along with the previous line.

tests/cross-language/go/.golangci.yml
```diff
@@ -0,0 +1,230 @@
...
+  # automatic updating of go.mod described above. Instead, it fails when any changes
+  # to go.mod are needed. This setting is most useful to check that go.mod does
+  # not need updates, such as in a continuous integration and testing system.
+  # If invoked with -mod=vendor, the go command assumes that the vendor
+  # directory holds the correct copies of dependencies and ignores
+  # the dependency descriptions in go.mod.
+  modules-download-mode: readonly
+
+  skip-dirs:
+    - cmd/example
```
aka-bo: can get rid of this

tests/cross-language/go/decrypt_test.go
aka-bo: It's definitely safe to reuse `err` here as their both usages are in the same scope so there's no shadowing. Just use the regular assignment operator (`err = ...`) instead of the short declaration operator (`err := ...`) and you'll be good.

tests/cross-language/java/scripts/build.sh
sushantmimani: we've used `mvn -U` in all compile scripts and some of them may possibly need to be changed as well. Will club this change along with those

PR #135, opened 2020-04-06
Asherah Server (Service Layer)

aka-bo:
This PR includes the initial service layer implementation 
* gRPC service definition
* golang grpc server
* sample client for python
* containerization samples for Docker Compose and Kubernetes


server/README.md
```diff
@@ -0,0 +1,131 @@
+# Asherah Server
+Asherah Server is intended to be used by those who wish to utilize application-layer encryption but are unable to take advantage of the SDK directly, e.g., Asherah lacks an implementation in their preferred programming language.
```
nikoo28, line 2: The lines in the README should end at 80 characters.

server/README.md
```diff
@@ -0,0 +1,131 @@
...
+## Overview
+Asherah Server is as a light-weight service layer built atop the Asherah SDK with encrypt/decrypt functionality exposed via a [gRPC](https://grpc.io) service. It uses a Unix domain socket for local inter-process communication and is designed to be deployed along side your application.
+
+To integrate with the service you will need to generate client interfaces from the [.proto](./protos/appencryption.proto) service definition. Detailed instructions for generating client code in a number of languages can be found in the [gRPC Quick Starts](https://grpc.io/docs/quickstart/).
+
+## Samples
+Full code for the client implementations used in the following samples, as well as the multi-container application configurations, can be found in the [samples](./samples) directory.
+
+### Docker Compose
+In this section we'll use Docker Compose to launch a multi-container application comprised of a simple Python client and an Asherah Server sidecar.
```
nikoo28, line 22: Can reword this a little bit. Get rid of `In this section`

server/README.md
```diff
@@ -0,0 +1,131 @@
...
+
+#### Prerequisites
+Ensure Docker Compose is installed on your local system. For installation instructions, see [Install Docker Compose](https://docs.docker.com/compose/install/).
+
+#### Build and run the sample application
+From the [samples](./samples) directory run `docker-compose up` which will launch the application defined in [docker-compose.yaml](./samples/docker-compose.yaml).
```
nikoo28, line 28: should this part be migrated to the README in samples directory?
As we are showing the `pwd` as `samples` in the below console snippet.

server/README.md
```diff
@@ -0,0 +1,131 @@
...
+myapp_1    | INFO:root:starting session for partitionid-1
+sidecar_1  | 2020/04/02 19:06:37 handling get-session for partitionid-1
+```
+
+At this point the sample client begins sending encrypt and decrypt messages to the server sidecar in a loop and you will see stream of log messages from both the client (myapp_1) and server (sidecar_1). Enter `CTRL+c` to shutdown the application.
+
+**NOTE**: Docker Compose will need to build the images for both containers the first time `docker-compose up` is run on your machine. This process typically takes between 5 to 10 minutes, but build times can vary considerably from one machine to another.
+
+### Kubernetes (kind)
+In this section we'll launch the the same multi-container application we used above but this time we'll use [kind](https://kind.sigs.k8s.io/) to spin up a local Kubernetes cluster and `kubectl` to launch the deployment.
```
nikoo28, line 47: Similarly here, can get rid of `in this section`

server/README.md
```diff
@@ -0,0 +1,131 @@
...
+INFO:root:starting test
+INFO:root:starting session for partitionid-1
+```
+
+And when finished, `kind delete cluster` can be used to delete the cluster.
+
+## Server Development
+
+### Code generation
+First you'll need to ensure the protobuf compiler is installed...
```
nikoo28, line 126: Similar to the above links in the readme, can also provide a link/instruction how to install protobuf compiler

server/README.md
```diff
@@ -0,0 +1,131 @@
...
+
+Then regenerate the gRPC server interfaces for go by running:
+```bash
+$ protoc --go_out=plugins=grpc:./go protos/appencryption.proto
+```
```
nikoo28, line 131: nit: missing extra line at the end of file

server/go/.dockerignore
```diff
@@ -0,0 +1,2 @@
+bin/
+cover.*
```
nikoo28, line 2: nit: missing extra line at the end of file

server/go/README.md
```diff
@@ -0,0 +1,30 @@
+# Asherah Server - Go
+
+## Running the server
+This assumes mysql is running on localhost and a preexisting asherah database
+
+```console
+$ go build -o server main.go
```
nikoo28, line 7: can edit the command line as `[user@machine go]$`

server/go/README.md
```diff
@@ -0,0 +1,30 @@
+# Asherah Server - Go
+
+## Running the server
+This assumes mysql is running on localhost and a preexisting asherah database
```
nikoo28, line 4: can point here a link to the `Metastore.md` doc file, which can help in setting up the metastore.

server/go/README.md
```diff
@@ -0,0 +1,30 @@
...
+    --product=servicelayer \
+    --expire-after=60m \
+    --check-interval=10m \
+    --metastore=rdbms \
+    --conn='root:my-secret-pw@tcp(0.0.0.0:3306)/asherah' \
+    --kms=static
+```
+
+Arguments can also be supplied using environment variables
+
```
nikoo28, line 19: be good to add a comment here stating the possible values for `ASHERAH_METASTORE_MODE` and `ASHERAH_KMS_MODE`

server/go/api/appencryption.pb.go
```diff
@@ -0,0 +1,776 @@
+// Code generated by protoc-gen-go. DO NOT EDIT.
```
nikoo28, line 1: do we need to commit this self generated code?

server/go/main.go
```diff
@@ -0,0 +1,95 @@
...
+		return
+	}
+
+	l, err := net.Listen("unix", string(opts.SocketFile))
+	if err != nil {
+		panic(err)
+	}
+	defer l.Close()
+
+	server := server.NewAppEncryption(opts.Asherah)
```
nikoo28, line 56: variable name `server` is same as the package name `server`. can rename it to something else

server/go/pkg/server/options.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"errors"
+	"strings"
+	"time"
+)
+
+//nolint:lll,staticcheck
+type Options struct {
+	ServiceName      string        `long:"service" required:"yes" description:"The name of this service" env:"ASHERAH_SERVICE_NAME"`
```
nikoo28, line 11: have been calling it ServiceId and ProductId at every place it seems. Better to decide on a consistent naming approach. Same with all the other options, I think we can decide on a consistent naming schema for all the available options for Java and Go servers

server/go/pkg/server/server.go
```diff
@@ -0,0 +1,295 @@
...
+
+		return persistence.NewDynamoDBMetastore(sess)
+	default:
+		return persistence.NewMemoryMetastore()
+	}
+}
+
+func NewKMS(opts *Options, crypto appencryption.AEAD) appencryption.KeyManagementService {
+	if opts.KMS == "static" {
+		kms, err := kms.NewStatic("thisistotallynotsecretdonotuse!!", aead.NewAES256GCM())
```
nikoo28, line 120: variable name `kms` same as package name. Is it fine? I see the IDE complain about it

server/go/pkg/server/server.go
```diff
@@ -0,0 +1,295 @@
...
+		panic(err)
+	}
+
+	return kms
+}
+
+func (s *streamer) Stream(stream pb.AppEncryption_SessionServer) error {
+	defer func() {
+		if s.handler != nil {
+			s.handler.Close()
```
nikoo28, line 139: missed error check with `Close`

server/protos/appencryption.proto
```diff
@@ -0,0 +1,75 @@
...
+  bytes data = 2;
+}
+
+message EnvelopeKeyRecord {
+  int64 created = 1;
+  bytes key = 2;
+  KeyMeta parent_key_meta = 3;
+}
+
+message KeyMeta {
```
nikoo28, line 48: how about calling this message as `ParentKeyMeta` instead of `KeyMeta`. When we call the apis, we usually do `EnvelopeKeyRecord.getKeyMeta`, but in a way we are getting the details of parent key meta. I believe `EnvelopeKeyRecord.getParentKeyMeta` would look more intuitive.

server/samples/clients/python/README.md
```diff
@@ -0,0 +1,33 @@
+# gRPC Client (Python) for Asherah Server
+A simple client application that demonstrates integrating with Asherah Server via a generated gRPC client.
+
+## Running the client
+Ensure the Asherah Server is running locally and listening on `unix:///tmp/appencryption.sock` and run:
+
+```console
+$ python appencryption_client.py
```
nikoo28, line 8: can use `[user@machine python]$`

server/samples/clients/python/README.md
```diff
@@ -0,0 +1,33 @@
...
+INFO:root:decrypting DRR
+INFO:root:received decrypted data: b'my "secret" data'
+INFO:root:test completed successfully
+```
+
+## (Re)generating the client
+Create a python virtual environment and install the client tools
+
+```console
+$ python3 -m venv .venv
```
nikoo28, line 22: ditto over here.

server/samples/metastore.sql
```diff
@@ -0,0 +1,9 @@
+CREATE DATABASE IF NOT EXISTS asherah;
```
nikoo28, line 1: we have the same DDL in `Metastore.md`. Can we somehow point at the same reference for both the places? This would help in removing redundancy.

server/README.md
aka-bo: There is no README in the samples directory and I think we're fine including this here for the time being. We can revisit if the scope of this doc balloons.

server/README.md
aka-bo: looks like it's there to me...

server/go/.dockerignore
aka-bo: same here... I re-cloned the repo for a sanity check.

server/go/api/appencryption.pb.go
aka-bo: Definitely. Committing the code makes for consistent, dependable builds and also drastically lowers the barrier of entry for those looking to try out the server.

server/go/pkg/server/options.go
aka-bo: Agreed. We need to introduce config consistency between implementations across the board... a unified configuration file, consistent flags, etc. My vote is to leave this as-is for now and revisit when we take up that effort.

server/go/pkg/server/server.go
aka-bo: Ignoring the error on deferred `Close` is a common practice. Just look at this as a `defer s.handler.Close()` with extra steps.

server/protos/appencryption.proto
nikoo28: cleared up the confusion in an offline discussion. resolving this.

server/samples/metastore.sql
aka-bo: This file is used by the `docker-compose-mysql.yaml` sample to init the mysql db and needs to be standalone.

server/README.md
sushantmimani: same. can see an extra line at EOF

server/go/main.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+
+		parser.WriteHelp(os.Stdout)
+
+		return
+	}
+
+	if err := validateOptions(opts); err != nil {
+		fmt.Println("Missing or invalid options:", err.Error())
+		fmt.Println()
```
sushantmimani, line 43: can possibly replace this with `fmt.Printf("Missing or invalid options: %s\n", err.Error())`?

server/go/pkg/server/database.go
```diff
@@ -0,0 +1,25 @@
...
+	}
+
+	dsn.ParseTime = true
+
+	db, err := sql.Open("mysql", dsn.FormatDSN())
+	if err != nil {
+		return nil, err
+	}
+
+	db.SetMaxOpenConns(90)
```
sushantmimani, line 22: should we add a note in the README about this config?

server/go/pkg/server/server.go
```diff
@@ -0,0 +1,295 @@
...
+	"log"
+
+	awssession "github.com/aws/aws-sdk-go/aws/session"
+	"github.com/godaddy/asherah/go/appencryption"
+	"github.com/godaddy/asherah/go/appencryption/pkg/crypto/aead"
+	"github.com/godaddy/asherah/go/appencryption/pkg/kms"
+	"github.com/godaddy/asherah/go/appencryption/pkg/persistence"
+	"github.com/godaddy/asherah/go/securememory/memguard"
+
+	pb "github.com/godaddy/asherah/server/go/api"
```
sushantmimani, line 15: any particular reason why we've renamed this import to pb and not just using `api`?

server/samples/clients/python/README.md
```diff
@@ -0,0 +1,37 @@
+# gRPC Client (Python) for Asherah Server
+A simple client application that demonstrates integrating with Asherah Server via a generated gRPC client.
+
+## Running the client
+Ensure the Asherah Server is running locally and listening on `unix:///tmp/appencryption.sock` and run:
```
sushantmimani, line 5: Should we add small sub-sections about running client in different modes - once, continuous, multiple clients ?

server/go/README.md
sushantmimani: possible values are explained in the `Configuring the server` section. should be fine if we don't add them here IMO

server/go/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+ENV GOOS=linux
+ENV GOARCH=amd64
+
+RUN go build -a -o /bin/asherah-server main.go
+
+FROM alpine:${ALPINEVERSION}
+WORKDIR /
+COPY --from=builder /bin/asherah-server .
+ENTRYPOINT ["/asherah-server"]
+CMD ["--help"]
```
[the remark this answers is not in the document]
aka-bo: I've found placing the `Dockerfile` in the context root to be a common practice, even in the case of multiple image variants where hyphenated suffixes are used, e.g., `Dockerfile-alpine`.

Is there a benefit to adopting the dir structure you referenced (other than consistency)? As long as we're consistent within the various roots we're probably ok.

server/go/README.md
```diff
@@ -0,0 +1,63 @@
...
+```bash
+export ASHERAH_SERVICE_NAME=example
+export ASHERAH_PRODUCT_NAME=servicelayer
+export ASHERAH_EXPIRE_AFTER=60m
+export ASHERAH_CHECK_INTERVAL=10m
+export ASHERAH_METASTORE_MODE=rdbms
+export ASHERAH_CONNECTION_STRING='root:my-secret-pw@tcp(0.0.0.0:3306)/asherah'
+export ASHERAH_KMS_MODE=static
+
+go run main -s /tmp/appencryption.sock
```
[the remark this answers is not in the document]
aka-bo: yep, good catch. will fix.

server/go/main.go
aka-bo: Two newlines (one blank line) is intentional. Replacing with the `fmt.Printf` example would eliminate the blank line. Of course, we could use `fmt.Printf` with `\n\n` but I tend to find `fmt.Println()` to be more explicit/readable.

server/go/pkg/server/database.go
aka-bo: I'm going to remove this line altogether as the default (unlimited) is a better choice.

server/go/pkg/server/server.go
aka-bo: Convention. The `pb` alias just calls attention to the fact that these types are provided by the protobuf code generator.

server/go/README.md
aka-bo: The `Configuring the server` section was added in response to @nikoo28 's suggestion :)

server/go/Dockerfile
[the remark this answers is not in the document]
sushantmimani: just for consistency. don't see an added benefit

nikoo28:
All concerns have been addressed. Just a few remaining documentation updates will be handled in another task when we finalize all the details with Java server.

PR #152, opened 2020-05-05
Release Go AppEncryption v0.1.0

sushantmimani:


nikoo28:
- Updating version file for AE
- Update AE to use SM v0.1.0

PR #162, opened 2020-05-07
Enable pr build

sushantmimani:
- Add a shell script to source env variables during build/test
- Update Java dependencies to use AE v0.1.0
- Update CircleCI config to use the shell script


sushantmimani:
Will create a new PR since this one isn't showing the build status

PR #163, opened 2020-05-07
Enable builds from forks

sushantmimani:
- Add a shell script to source env variables during build/test
- Update Java dependencies to use AE v0.1.0
- Update CircleCI config to use the shell script


.circleci/config.yml
```diff
@@ -3,6 +3,17 @@ version: 2.1
...
+parameters:
+  db-name:
+    type: "string"
+    default: "testdb"
+  db-user:
+    type: "string"
+    default: "root"
+  db-password:
+    type: "string"
+    default: "Password123"
```
aka-bo, line 15: Looks like these parameters are duplicated in the new `set_env_variables.sh` script. Could this new script be defined here in the `config.yml` instead to eliminate duplication?

.circleci/config.yml
```diff
@@ -515,7 +530,7 @@ jobs:
           name: Build
           command: |
             cd samples/go/referenceapp
-            go get github.com/godaddy/asherah/go/appencryption@${CIRCLE_BRANCH}
+            go mod edit -replace github.com/godaddy/asherah/go/appencryption=../../../go/appencryption
```
aka-bo, line 533: Should update the referenceapp's `go.mod` to reference to latest version instead of using replace. This would be consistent with the java pom updates as well.

sushantmimani: referenceapp should ideally use the current AE version, because no AE updates should break the referenceapp. 

The jar for java and the nuget package for C# are locally installed, so we can make that happen but wasn't able to find a similar method for Go.

aka-bo: > referenceapp (and possibly cltf) should ideally use the current AE version, because no AE updates should break the referenceapp.

I think this holds true for the cross-language tests but not so much for the reference apps. The cltf aims to ensure changes in one implementation don't impact cross-language compatibility, so always testing against the latest AE (current branch) makes sense because we want discover issues asap, prior to releasing a new version.

The reference apps, however, serve a different purpose, they exist as a guide for developers looking to use the SDK. And because users of the SDK are encouraged to use the latest released version (as opposed to `master`), our CI pipeline should also follow suit.

.circleci/config.yml
```diff
@@ -57,10 +67,8 @@ workflows:
             - csharp-app-encryption
       - go-securememory
       - go-appencryption
-      - go-referenceapp:
-          requires:
-            - go-appencryption
-      - cross-language:
+      - samples-go-reference-app
```
nikoo28, line 70: nit: good callout on renaming it. but it should be `samples-go-referenceapp`

sushantmimani: followed the same naming convention as the jobs for Java and C#. seemed better to keep the naming scheme consistent within the config file

PR #167, opened 2020-05-12
CLTF framework implementation for Go and Java sidecar

nikoo28:
- use java 1.8 in cross language tests
- add cltf cache
- encrypt and decrypt using both the side cars
- java and go sidecars now return consistent results
- required changes in test classes
- made the static kms key consistent
- purge old circleci cache and made names consistent
- added back dependencies as caches are progressive
- fixed linting of go cltf


tests/cross-language/scripts/encrypt_all.sh
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env bash
+set -e
+
+export PATH=$GOPATH/bin:$GOROOT/bin:$PATH
+
+export ASHERAH_SERVICE_NAME=service
+export ASHERAH_PRODUCT_NAME=product
+export ASHERAH_KMS_MODE=static
```
sushantmimani, line 8: same as above

tests/cross-language/scripts/decrypt_all.sh
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env bash
+set -e
+
+export PATH=$GOPATH/bin:$GOROOT/bin:$PATH
+
+export ASHERAH_SERVICE_NAME=service
+export ASHERAH_PRODUCT_NAME=product
+export ASHERAH_KMS_MODE=static
```
sushantmimani, line 8: since these are used in both scripts, we should move them to the config file to avoid duplication

samples/go/referenceapp/main.go
```diff
@@ -56,7 +56,7 @@ func connectSQL() error {
 
 func createMetastore() appencryption.Metastore {
 	switch opts.Metastore {
-	case "SQL":
+	case "RDBMS":
 		logger.Info("using sql metastore")
```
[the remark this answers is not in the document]
nikoo28: discussed on call, leaving as is. True to the documentation.

PR #171, opened 2020-05-14
Codecov exclude files

sushantmimani:
- Rename `.codecov.yml` to `codecov.yml`
- Ignore autogenerated and server startup files
- Add `conf: codecov.yml` to codecov upload step in circleci config
- Update codecov.yml as per [suggestion](https://community.codecov.io/t/ignore-option-not-working/1417) from codecov support (


nikoo28:
discussed and made changes in the latest commit. looks good now 🦅

PR #188, opened 2020-05-19
Non-Root PHP Sample Client

jeffchannell:
This PR adds an additional sample client in PHP as an example of running Asherah as a non-root user.


aka-bo:
Thank you for the contribution! Unfortunately, we are not accepting new sample applications at this time. Apologies for the delay in closing your PR and thanks again for taking the time to create this sample application.

PR #194, opened 2020-05-20
Removing "GoDaddy.Asherah.Logging"

sjlbos:
This change removes the `GoDaddy.Asherah.Logging` dependency from the `GoDaddy.Asherah.PlatformNative` project. This is
the first step of several to remove this dependency from `GoDaddy.Asherah.AppEncryption`.


csharp/SecureMemory/SecureMemory/ProtectedMemoryImpl/Libc/LibcProtectedMemoryAllocatorLP64.cs
```diff
@@ -51,13 +48,6 @@ public void SetReadWriteAccess(IntPtr pointer, ulong length)
         // ************************************
         public virtual IntPtr Alloc(ulong length)
         {
-#pragma warning disable 162
-            if (Debug.On)
-            {
-                Logger.LogDebug("attempting to alloc length {length}", length);
-            }
-#pragma warning restore 162
-
```
sushantmimani, line 60: We don't want to remove logging entirely. We have decided that we should use NLog for logging purposes.
https://github.com/sushantmimani/asherah/blob/d9b244d48e572dd0d3e10789d6911eafc28ddb7f/csharp/SecureMemory/SecureMemory/ProtectedMemoryImpl/Libc/LibcProtectedMemoryAllocatorLP64.cs#L14

csharp/SecureMemory/SecureMemory/ProtectedMemoryImpl/ProtectedMemorySecret.cs
```diff
@@ -35,12 +31,6 @@ internal ProtectedMemorySecret(byte[] sourceBytes, IProtectedMemoryAllocator all
             {
                 throw new ProtectedMemoryAllocationFailedException("Protected memory allocation failed");
             }
-            #pragma warning disable 162
-            if (Debug.On)
-            {
-                Logger.LogDebug("allocated: {pointer}", pointer);
-            }
-            #pragma warning restore 162
```
sushantmimani, line 43: same as above

csharp/SecureMemory/SecureMemory/ProtectedMemoryImpl/ProtectedMemorySecret.cs
```diff
@@ -84,13 +74,6 @@ public override TResult WithSecretBytes<TResult>(Func<byte[], TResult> funcWithS
                 SetReadAccessIfNeeded();
                 try
                 {
-                    #pragma warning disable 162
-                    if (Debug.On)
-                    {
-                        Logger.LogDebug("reading: {pointer}", pointer);
-                    }
-                    #pragma warning restore 162
-
```
sushantmimani, line 93: same as above

csharp/SecureMemory/SecureMemory/ProtectedMemoryImpl/Libc/LibcProtectedMemoryAllocatorLP64.cs
sjlbos: Because this is a library, we shouldn't dictate a logging framework but should instead use .NET's ILogger interface and inject an appropriate logging implementation. I can add this, but it's going to impact the existing constructors. Since this logging is so low-level, I felt it better to just remove this stuff entirely, since this information can be obtained via a debugger (I would never want to see this logged in a deployed environment, since there are better ways to gain insight into things like memory allocation, ie using a profiler). If you still feel strongly about this, I can add `ILoggerFactory` as a constructor param, but it will make the library more cumbersome.

aka-bo: @sjlbos I'm not opposed to the removal of logging in this specific library, because - as you pointed out - it's all very debug/troubleshooting centric, plus it's all gated behind a hardcoded `Debug.On` value anyways. However, this doesn't hold true with the AppEncryption library, as a significant portion of those logs expose warning/error information which _is_ critically important in a deployed environment and unavailable via other means. So, admittedly, I'm more concerned with the general approach of this PR (i.e. remove all logging) and what it implies for the follow up PRs alluded to in this PR's summary than I am with the PR content itself.

That said, I have no qualms with approving this PR as-is with the caveat that follow-up PRs will need to take a different approach (e.g. injection of an ILogger) to retain the logging capability supported today.

sjlbos: Agreed, that was going to be my approach (updating AppEncryption to use `ILoggerFactory`). I wanted to remove this transitive dependency first so that it was easier to clean up the legacy logging in `AppEncryption`.

aka-bo: Approving.

@sushantmimani I'll leave final approval and merge up to you (assuming you agree with the comments above of course)

sushantmimani: Sounds good to me. Approving.

PR #199, opened 2020-05-22
Remove local install

sushantmimani:
- Remove local install for AE and SM
- Delete local_install scripts
- Update cache to v3
- Use `ProjectReference` instead of `PackageReference` in C#  CLTF
- Use `${revision}` to pass the AE version in  Java CLTF
- Use dotnet core 2.1 docker images


.circleci/config.yml
```diff
@@ -596,12 +571,6 @@ jobs:
             apt-get update
             apt-get -y upgrade
             apt-get install -y ca-certificates wget curl build-essential
-      - restore_cache:
-          name: Restore clients packages
-          keys:
-            - v2-clients-{{ .Branch }}-{{ .BuildNum }}
-            - v2-clients-{{ .Branch }}-
-            - v2-clients-
```
nikoo28, line 604: did these get removed accidentally?

sushantmimani: yep, corrected them now.

PR #202, opened 2020-05-28
Java: Add global table support

sushantmimani:


java/app-encryption/src/main/java/com/godaddy/asherah/appencryption/Partition.java
```diff
@@ -1,40 +1,50 @@
...
+   * Gets the SystemKey Id associated with the partition
+   * @return The System Key Id
+   */
+  public abstract String getSystemKeyId();
+
+  /**
+   * Gets the Intermediate Key Id associated with the partition
+   * @return The IntermediateKey Id
+   */
+  public abstract String getIntermediateKeyId();
```
nikoo28, line 49: get consistent on the javadoc comment. `SystemKey Id` or `System Key Id` as written in intermediate key?

java/app-encryption/src/main/java/com/godaddy/asherah/appencryption/Partition.java
```diff
@@ -1,40 +1,50 @@
...
+  /**
+   * Gets the serviceId for a partition
+   * @return  The service id
+   */
   String getServiceId() {
     return serviceId;
   }
 
+  /**
+   * Gets the productId for a partition
```
nikoo28, line 32: for all the description javadoc, i think we can remove the word `partition` since the classname is `Partition`. So when you say `getServiceId`, it is implied that it is a service id for the partition.

Given that, we may be also able to remove the descriptions all together. Just return comments should be self explanatory.

java/app-encryption/src/main/java/com/godaddy/asherah/appencryption/SuffixedPartition.java
```diff
@@ -0,0 +1,28 @@
...
+  public String getIntermediateKeyId() {
+    return "_IK_" + getPartitionId() + "_" + getServiceId() + "_" + getProductId() + "_" + this.regionSuffix;
+  }
+
+  @Override
+  public String toString() {
+    return getClass().getSimpleName() + "[partitionId=" + getPartitionId() +
+      ", serviceId=" + getServiceId() + ", productId=" + getProductId() + ", regionSuffix=" + this.regionSuffix + "]";
+  }
+}
```
nikoo28, line 28: one approach you can take is having the default implementations in the abstract class `Partition`.

The `DefaultPartition` class would then just reduce to a constructor with default implementations.
The `SuffixedPartition` class would then call the default implementations and just add ```"_" + this.regionSuffix;```

java/app-encryption/src/main/java/com/godaddy/asherah/appencryption/persistence/InMemoryMetastoreImpl.java
```diff
@@ -41,4 +41,9 @@ public boolean store(final String keyId, final Instant created, final T value) {
       return true;
     }
   }
+
+  @Override
+  public String getRegionSuffix() {
+    return "";
+  }
```
nikoo28, line 48: can add this default implementation to the `Metastore` interface. Would help in reducing a lot of redundant code.

java/app-encryption/src/main/java/com/godaddy/asherah/appencryption/DefaultPartition.java
```diff
@@ -0,0 +1,24 @@
...
+  @Override
+  public String getIntermediateKeyId() {
+    return "_IK_" + getPartitionId() + "_" + getServiceId() + "_" + getProductId();
+  }
+
+  @Override
+  public String toString() {
+    return getClass().getSimpleName() + "[partitionId=" + getPartitionId() +
+      ", serviceId=" + getServiceId() + ", productId=" + getProductId() + "]";
+  }
```
nikoo28, line 23: this can be avoided.

java/app-encryption/src/main/java/com/godaddy/asherah/appencryption/persistence/JdbcMetastoreImpl.java
```diff
@@ -130,6 +130,11 @@ public boolean store(final String key, final Instant created, final JSONObject v
     });
   }
 
+  @Override
+  public String getRegionSuffix() {
+    return "";
+  }
+
```
nikoo28, line 137: this would be no longer needed.

java/app-encryption/src/main/java/com/godaddy/asherah/appencryption/persistence/Metastore.java
```diff
@@ -30,4 +30,9 @@
    */
   boolean store(String keyId, Instant created, V value);
 
+  /**
+   * Returns the region suffix for the metastore or "" if region suffix option is disabled
```
nikoo28, line 34: description can be just `Returns the region suffix or "" if region suffix option is disabled`

java/app-encryption/src/test/java/com/godaddy/asherah/appencryption/SuffixedPartitionTest.java
```diff
@@ -0,0 +1,42 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class SuffixedPartitionTest {
+
+    private final static String testPartitionId = "test_partition_id";
+    private final static String testServiceId = "test_service_id";
+    private final static String testProductId = "test_product_id";
+    private final static String testRegionSuffix = "test_region_suffix";
+
+
```
nikoo28, line 15: nit: remove extra line

tests/cross-language/scripts/decrypt_all.sh
```diff
@@ -43,7 +43,7 @@ export ASHERAH_EXPIRE_AFTER=90
 export ASHERAH_CHECK_INTERVAL=10
 export ASHERAH_METASTORE_MODE=jdbc
 export ASHERAH_CONNECTION_STRING='jdbc:mysql://127.0.0.1:3306/testdb?user=root&password=Password123'
-java -jar /home/circleci/.m2/repository/com/godaddy/asherah/grpc-server/1.0.0-SNAPSHOT/*dependencies.jar &
+java -jar ~/.m2/repository/com/godaddy/asherah/grpc-server/1.0.0-SNAPSHOT/*dependencies.jar &
```
nikoo28, line 46: this looks invalid. It seems that we are looking for a jar in `~/.m2`. There wouldn't be an updated jar in the cache since you removed the `restore java cache` option

nikoo28: can discuss the different approaches offline on how to tackle this

java/app-encryption/src/main/java/com/godaddy/asherah/appencryption/SuffixedPartition.java
sushantmimani: Personally, I like having separate implementations better. It keeps things simpler. Any  particular advantage that you foresee using the above approach?

java/app-encryption/src/main/java/com/godaddy/asherah/appencryption/persistence/InMemoryMetastoreImpl.java
sushantmimani: Having separate implementations is better in terms of extensibility if at  some point we decide to enable region suffix for the JDBC version. That being said, I am ok changing this to a `default` method if you feel too strongly about  it.

tests/cross-language/scripts/decrypt_all.sh
sushantmimani: the jar will be pushed to ~/.m2 every time we build the Java CLTF https://github.com/sushantmimani/asherah/blob/62f97205bc69157313c53db8bcdce5d1b8722f37/.circleci/config.yml#L685

java/app-encryption/src/main/java/com/godaddy/asherah/appencryption/DefaultPartition.java
sushantmimani: refer comments above. this would move from DefaultPartition to  Partition so not truly  avoided.

java/app-encryption/src/test/java/com/godaddy/asherah/appencryption/SessionFactoryTest.java
```diff
@@ -85,6 +86,8 @@ void testConstructor() {
 
   @Test
   void testSessionCacheSetupAndClose() {
+    when(metastoreMock.getRegionSuffix()).thenReturn("");
```
nikoo28, line 89: can move this mock setup to the `@BeforeEach` annotation. can help to get rid of some redundant code.

java/app-encryption/src/main/java/com/godaddy/asherah/appencryption/SuffixedPartition.java
nikoo28: this has been taken care of

java/app-encryption/src/main/java/com/godaddy/asherah/appencryption/persistence/InMemoryMetastoreImpl.java
nikoo28: discussed about this, we can override method as/when required in JDBC version.

tests/cross-language/scripts/decrypt_all.sh
nikoo28: discussed an approach where we get rid of local install completely. This makes sure that we are always using the build from the current branch/tag.

java/app-encryption/src/main/java/com/godaddy/asherah/appencryption/persistence/Metastore.java
```diff
@@ -30,4 +30,11 @@
    */
   boolean store(String keyId, Instant created, V value);
 
+  /**
+   * Returns the region suffix or "" if region suffix option is disabled
+   * @return The region suffix
+   */
+  default String getRegionSuffix() {
+    return "";
+  }
```
aka-bo, line 39: Hmm, modifying the metastore interface was something I intentionally avoided with the Go implementation, though I suppose that's not as easy to avoid here. Shouldn't be an issue as the default method ensures backward compatibility... just thinking out loud :)

java/app-encryption/src/main/java/com/godaddy/asherah/appencryption/persistence/DynamoDbMetastoreImpl.java
```diff
@@ -42,13 +42,15 @@
...
   // Table instance can be cached since thread-safe and no state other than description, which we don't use
   private final Table table;
+  private final String regionSuffix;
 
   public static Builder newBuilder() {
     return new Builder();
   }
 
-  DynamoDbMetastoreImpl(final DynamoDB dynamoDbDocumentClient) {
+  DynamoDbMetastoreImpl(final DynamoDB dynamoDbDocumentClient, final String region) {
```
aka-bo, line 51: new argument should be renamed to avoid confusion, i.e. `regionSuffix` or `suffix` as opposed to `region`

PR #203, opened 2020-05-29
Updated C# metastore exception handling

sjlbos:
This change modifies the existing C# `IMetastore` implementations so that exceptions that get raised by the database
driver are not suppressed and logged, but instead, bubble up to the caller. The reason for this change is two-fold:

1. Remove unnecessary logging deep in the call stack: by removing the `try`/`catch` blocks in the metastore
implementations, it removes the need for logging at the bottom of the stack, which will make future changes to remove
the existing logging implementation simpler.

2. It preserves the stack trace/exception so that the low-level exception can get added as the `InnerException` property
value of the exception that eventually gets returned by the library. This will make it easier for developers to diagnose
problems with their metastore configurations, since they will now only see a single exception in their logging with a
clear cause and effect between the low-level exception and the exception that gets thrown by the library.


sushantmimani:
Closing this PR due to no response from the contributor.

PR #207, opened 2020-06-03
Release Java AE v0.1.1

sushantmimani:
- Add Global Tables support for DynamoDb


sushantmimani:
Discovered a bug in the java global table support. will re-open after bug fix

PR #210, opened 2020-06-04
JAVA: DynamoDb config updates

sushantmimani:
- Fix backwards compatibility bug for region suffixes
- Add endpoint configuration builder step
- Add configurable region builder step
- Add configurable table name builder step
- Add dynamodb backward compatibility test in the test-app

### DIFF
--- a/java/app-encryption/pom.xml
+++ b/java/app-encryption/pom.xml
@@ -4,7 +4,7 @@
 
   <groupId>com.godaddy.asherah</groupId>
   <artifactId>appencryption</artifactId>
-  <version>0.1.0</version>
+  <version>0.1.1-alpha</version>
   <name>Asherah</name>
   <description>
     An application-layer encryption SDK that provides advanced encryption features and in depth defense against
--- a/java/app-encryption/pom.xml
+++ b/java/app-encryption/pom.xml
@@ -4,7 +4,7 @@
 
   <groupId>com.godaddy.asherah</groupId>
   <artifactId>appencryption</artifactId>
-  <version>0.1.1-alpha</version>
+  <version>0.1.0</version>
   <name>Asherah</name>
   <description>
     An application-layer encryption SDK that provides advanced encryption features and in depth defense against
--- a/java/app-encryption/src/main/java/com/godaddy/asherah/appencryption/persistence/DynamoDbMetastoreImpl.java
+++ b/java/app-encryption/src/main/java/com/godaddy/asherah/appencryption/persistence/DynamoDbMetastoreImpl.java
@@ -157,7 +157,7 @@ public class DynamoDbMetastoreImpl implements Metastore<JSONObject> {
 
     private AmazonDynamoDB client;
 
-    private AmazonDynamoDBClientBuilder standardBuilder = AmazonDynamoDBClientBuilder.standard();
+    private final AmazonDynamoDBClientBuilder standardBuilder = AmazonDynamoDBClientBuilder.standard();
     private String regionSuffix = DEFAULT_REGION_SUFFIX;
     private String tableName = DEFAULT_TABLE_NAME;
 
@@ -205,7 +205,7 @@ public class DynamoDbMetastoreImpl implements Metastore<JSONObject> {
   public interface RegionStep {
     /**
      * Specifies the region for the AWS DynamoDb client
-     * @param region
+     * @param region The region for the DynamoDb client
      * @return The current {@code BuildStep} instance
      */
     BuildStep withRegion(String region);
--- a/java/app-encryption/src/main/java/com/godaddy/asherah/appencryption/persistence/DynamoDbMetastoreImpl.java
+++ b/java/app-encryption/src/main/java/com/godaddy/asherah/appencryption/persistence/DynamoDbMetastoreImpl.java
@@ -43,7 +43,6 @@ public class DynamoDbMetastoreImpl implements Metastore<JSONObject> {
   // Table instance can be cached since thread-safe and no state other than description, which we don't use
   private final Table table;
   private final DynamoDB client;
-
   private final String tableName;
   private final String regionSuffix;
 
--- a/java/app-encryption/src/main/java/com/godaddy/asherah/appencryption/persistence/DynamoDbMetastoreImpl.java
+++ b/java/app-encryption/src/main/java/com/godaddy/asherah/appencryption/persistence/DynamoDbMetastoreImpl.java
@@ -152,13 +152,14 @@ public class DynamoDbMetastoreImpl implements Metastore<JSONObject> {
   }
 
   public static final class Builder implements BuildStep, EndPointStep, RegionStep {
+    static final String DEFAULT_TABLE_NAME = "EncryptionKey";
+    static final String DEFAULT_REGION_SUFFIX = "";
 
     private AmazonDynamoDB client;
 
-    private String regionSuffix = "";
     private AmazonDynamoDBClientBuilder standardBuilder = AmazonDynamoDBClientBuilder.standard();
-    private String tableName = "EncryptionKey";
-
+    private String regionSuffix = DEFAULT_REGION_SUFFIX;
+    private String tableName = DEFAULT_TABLE_NAME;
 
     @Override
     public BuildStep withEndPointConfiguration(final String endPoint, final String signingRegion) {
--- a/java/app-encryption/src/main/java/com/godaddy/asherah/appencryption/persistence/DynamoDbMetastoreImpl.java
+++ b/java/app-encryption/src/main/java/com/godaddy/asherah/appencryption/persistence/DynamoDbMetastoreImpl.java
@@ -40,11 +40,11 @@ public class DynamoDbMetastoreImpl implements Metastore<JSONObject> {
       Metrics.timer(MetricsUtil.AEL_METRICS_PREFIX + ".metastore.dynamodb.loadlatest");
   private final Timer storeTimer = Metrics.timer(MetricsUtil.AEL_METRICS_PREFIX + ".metastore.dynamodb.store");
 
-  // Table instance can be cached since thread-safe and no state other than description, which we don't use
-  private final Table table;
   private final DynamoDB client;
   private final String tableName;
   private final String regionSuffix;
+  // Table instance can be cached since thread-safe and no state other than description, which we don't use
+  private final Table table;
 
   public static Builder newBuilder() {
     return new Builder();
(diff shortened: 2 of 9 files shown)


java/app-encryption/pom.xml
```diff
@@ -4,7 +4,7 @@
 
   <groupId>com.godaddy.asherah</groupId>
   <artifactId>appencryption</artifactId>
-  <version>0.1.1-alpha</version>
+  <version>0.1.0</version>
```
aka-bo, line 7: Why the version _decrement_?

sushantmimani: ugh! wrong POM update

java/app-encryption/src/main/java/com/godaddy/asherah/appencryption/persistence/DynamoDbMetastoreImpl.java
```diff
@@ -139,31 +143,87 @@ public String getRegionSuffix() {
...
+  public static final class Builder implements BuildStep, EndPointStep, RegionStep {
 
-    private String regionSuffix = "";
     private AmazonDynamoDB client;
 
+    private String regionSuffix = "";
+    private AmazonDynamoDBClientBuilder standardBuilder = AmazonDynamoDBClientBuilder.standard();
+    private String tableName = "EncryptionKey";
+
+
```
nikoo28, line 162: nit: extra line

java/app-encryption/src/main/java/com/godaddy/asherah/appencryption/persistence/DynamoDbMetastoreImpl.java
```diff
@@ -139,31 +143,87 @@ public String getRegionSuffix() {
...
+  DynamoDB getClient() {
+    return client;
+  }
+
+  public static final class Builder implements BuildStep, EndPointStep, RegionStep {
 
-    private String regionSuffix = "";
     private AmazonDynamoDB client;
 
+    private String regionSuffix = "";
```
nikoo28, line 158: nit: 2 extra lines above can be removed

java/app-encryption/src/main/java/com/godaddy/asherah/appencryption/persistence/DynamoDbMetastoreImpl.java
```diff
@@ -139,31 +143,87 @@ public String getRegionSuffix() {
...
+     */
+    BuildStep withDynamoDbRegionSuffix(String suffix);
+
+    /**
+     * Specifies the name of the table
+     * @param table The name of the table
      * @return The current {@code BuildStep} instance.
      */
-    BuildStep withDynamoDbRegionSuffix(String region);
+    BuildStep withTableName(String table);
```
nikoo28, line 226: this step should be modified to use kind of a step builder.
currently, i can create a `DynamoDbMetastoreImpl` without specifying a table name, hence causing an error.

Having a step builder would force you to specify a table name before calling the actual `build()` step.

java/app-encryption/src/test/java/com/godaddy/asherah/appencryption/persistence/DynamoDbMetastoreImplTest.java
```diff
@@ -216,17 +218,54 @@ void testStoreWithFailureShouldThrowException() {
...
+    DynamoDbMetastoreImpl dynamoDbMetastore = DynamoDbMetastoreImpl.newBuilder()
+      .withDynamoDbRegionSuffix("us-west-2")
+      .build();
 
     assertEquals("", dynamoDbMetastoreImpl.getRegionSuffix());
     assertEquals("us-west-2", dynamoDbMetastore.getRegionSuffix());
   }
+
+  @Test
+  void testWithTableName() {
```
nikoo28, line 244: The name of this test can be changed to something meaningful. A lot is happening in this test.

I believe this test creates a client using standard builder provided by AWS and then you create a table in it.
Later you verify that the library is able to connect to the same table, which got created in the previous step.

tests/java/test-app/pom.xml
```diff
@@ -292,6 +319,24 @@
               </buildArgs>
             </configuration>
           </plugin>
+          <plugin>
```
nikoo28, line 322: this can also be commented out.

java/app-encryption/src/main/java/com/godaddy/asherah/appencryption/persistence/DynamoDbMetastoreImpl.java
sushantmimani: Creating a `DynamoDbMetastoreImpl` object without a table name does not cause an error.
The table name defaults with `EncryptionKey` if the `withTableName` builder step isn't  called. https://github.com/sushantmimani/asherah/blob/866d6ccc20019257494bed33b04e158484be8f98/java/app-encryption/src/main/java/com/godaddy/asherah/appencryption/persistence/DynamoDbMetastoreImpl.java#L160

All our existing tests run without this step
https://github.com/sushantmimani/asherah/blob/866d6ccc20019257494bed33b04e158484be8f98/java/app-encryption/src/test/java/com/godaddy/asherah/appencryption/persistence/DynamoDbMetastoreImplTest.java#L73

sushantmimani: Updated the code to be more specific regarding default values

java/app-encryption/src/test/java/com/godaddy/asherah/appencryption/persistence/DynamoDbMetastoreImplTest.java
```diff
@@ -216,17 +218,54 @@ void testStoreWithFailureShouldThrowException() {
...
 
-    DynamoDbMetastoreImpl dynamoDbMetastoreImpl =
-      DynamoDbMetastoreImpl.newBuilder().build();
     assertNotNull(dynamoDbMetastoreImpl);
   }
 
   @Test
   void testMetastoreWithRegionSuffix() {
-    DynamoDbMetastoreImpl dynamoDbMetastore = DynamoDbMetastoreImpl.newBuilder().withDynamoDbRegionSuffix("us-west-2").build();
+    DynamoDbMetastoreImpl dynamoDbMetastore = DynamoDbMetastoreImpl.newBuilder()
```
nikoo28, line 235: the above testMethod should be renamed to `testBuilderPathWithRegionSuffix`

java/app-encryption/src/test/java/com/godaddy/asherah/appencryption/persistence/DynamoDbMetastoreImplTest.java
```diff
@@ -216,17 +218,54 @@ void testStoreWithFailureShouldThrowException() {
...
+      .withDynamoDbRegionSuffix("us-west-2")
+      .build();
 
     assertEquals("", dynamoDbMetastoreImpl.getRegionSuffix());
     assertEquals("us-west-2", dynamoDbMetastore.getRegionSuffix());
   }
+
+  @Test
+  void testMetastoreWithConfigurableTableName() {
```
nikoo28, line 244: following a similar convention this can be renamed to `testBuilderPathWithTableName`

java/app-encryption/src/test/java/com/godaddy/asherah/appencryption/persistence/DynamoDbMetastoreImplTest.java
```diff
@@ -216,17 +218,54 @@ void testStoreWithFailureShouldThrowException() {
...
+
+    DynamoDbMetastoreImpl dynamoDbMetastore = DynamoDbMetastoreImpl.newBuilder()
+      .withEndPointConfiguration("http://localhost:8000", "us-west-2")
+      .withTableName(tableName)
+      .build();
+    Optional<JSONObject> actualJsonObject = dynamoDbMetastore.load(TEST_KEY, instant);
+
+    assertTrue(actualJsonObject.isPresent());
+    assertEquals(keyRecord, actualJsonObject.get().toMap());
+  }
```
nikoo28, line 270: should we add one more test something like `testBuilderPathWithEndpointURL` because that is also a sufficient condition to build the metastore object and it'd be good to verify that it works correctly.

tests/java/test-app/src/main/java/com/godaddy/asherah/testapp/regression/DynamoDbGlobalTableTest.java
```diff
@@ -0,0 +1,127 @@
...
+//        originalPayloadString = "mysupersecretpayload";
+//
+//        byte[] dataRowRecordBytes =
+//          sessionBytes.encrypt(originalPayloadString.getBytes(StandardCharsets.UTF_8));
+//
+//        // Consider this us "persisting" the DRR
+//        dataRowString = Base64.getEncoder().encodeToString(dataRowRecordBytes);
+//      }
+//    }
+//
```
nikoo28, line 102: Can add a comment here, about what we are trying to test.
Something along the lines of "to ensure backward compatibility, we try to encrypt a record without a suffix, and then decrypt it back with a region suffix"
Can also add that this test might be removed in future with 1.0.0 release

java/app-encryption/src/test/java/com/godaddy/asherah/appencryption/persistence/DynamoDbMetastoreImplTest.java
sushantmimani: The test itself run using the `withEndPoint` builder step so i think that covers that aspect

sushantmimani: All DynamoDb test setup is done using just the `withEndPointConfiguration` builder step. Do you still want to have an additional test?

nikoo28: upon a second thought it seems un-necessary...but just in a future case if we have to modify the setup method in some scenario, then this test would ensure that the builder path works correctly.

sushantmimani: Adding to testing in isolation

tests/java/test-app/src/main/java/com/godaddy/asherah/testapp/regression/DynamoDbGlobalTableTest.java
```diff
@@ -0,0 +1,127 @@
+/*
+Currently disabled because DynamoDbLocal isn't playing nice with maven-shade-plugin.
```
aka-bo, line 2: Could extend this comment to include a `TODO` regarding the "placeholder" nature of the tests and the plans to relocate the tests in general

tests/java/test-app/pom.xml
```diff
@@ -292,6 +319,25 @@
               </buildArgs>
             </configuration>
           </plugin>
+        <!--    Enable to run the DynamoDbGlobalTableTest-->
+<!--          <plugin>-->
```
aka-bo, line 323: nit: whitespace

PR #217, opened 2020-06-06
[Go] DynamoDB config updates

aka-bo:
- [server/go] add `--dynamodb-endpoint` flag
- [server/go] add `--dynamodb-region` flag
- [go] WithTableName option for DynamoDBMetastore

### DIFF
--- a/go/appencryption/pkg/persistence/dynamodb_test.go
+++ b/go/appencryption/pkg/persistence/dynamodb_test.go
@@ -34,6 +34,7 @@ type DynamoDBSuite struct {
 }
 
 const (
+	tableName            = "CustomTableName"
 	portProtocolDynamoDB = "8000/tcp"
 	maxTriesDynamoDB     = 5
 	waitTimeDynamoDB     = 10
@@ -172,7 +173,7 @@ func (suite *DynamoDBSuite) SetupTest() {
 	}
 	suite.putItemInDynamoDB(getDynamoDBItem(en, suite.instant))
 
-	suite.dynamodbMetastore = NewDynamoDBMetastore(suite.sess)
+	suite.dynamodbMetastore = NewDynamoDBMetastore(suite.sess, WithTableName(tableName))
 	suite.prefixedDynamodbMetastore = NewDynamoDBMetastore(suite.sess, WithDynamoDBRegionSuffix(true))
 }
 
@@ -339,10 +340,23 @@ func (suite *DynamoDBSuite) TestDynamoDBMetastore_Store_WithFailureShouldReturnE
 
 func (suite *DynamoDBSuite) TestDynamoDBMetastore_WithDynamoDBRegionSuffix() {
 	// keyPrefix should be empty unless WithDynamoDBRegionSuffix is used
-	assert.Empty(suite.T(), suite.dynamodbMetastore.keySuffix)
+	assert.Empty(suite.T(), suite.dynamodbMetastore.GetSuffix())
 
 	// WithDynamoDBRegionSuffix should set the keyPrefix equal to the client's region
-	assert.Equal(suite.T(), *suite.sess.Config.Region, suite.prefixedDynamodbMetastore.keySuffix)
+	assert.Equal(suite.T(), *suite.sess.Config.Region, suite.prefixedDynamodbMetastore.GetSuffix())
+}
+
+func (suite *DynamoDBSuite) TestDynamoDBMetastore_WithTableName() {
+	table := "DummyTable"
+	db := NewDynamoDBMetastore(suite.sess, WithTableName(table))
+
+	assert.Equal(suite.T(), table, db.tableName)
+}
+
+func (suite *DynamoDBSuite) TestDynamoDBMetastore_DefaultTableName() {
+	db := NewDynamoDBMetastore(suite.sess)
+
+	assert.Equal(suite.T(), defaultTableName, db.tableName)
 }
 
 func TestDynamoSuite(t *testing.T) {
--- a/server/go/README.md
+++ b/server/go/README.md
@@ -56,6 +56,8 @@ Asherah Options:
                                           --metastore=dynamodb) [$ASHERAH_ENABLE_REGION_SUFFIX]
       --dynamodb-endpoint=                An optional endpoint URL (hostname only or fully qualified URI) (only
                                           supported by --metastore=dynamodb) [$ASHERAH_DYNAMODB_ENDPOINT]
+      --dynamodb-region=                  The AWS region for DynamoDB requests (defaults to globally configured region)
+                                          (only supported by --metastore=dynamodb) [$ASHERAH_DYNAMODB_REGION]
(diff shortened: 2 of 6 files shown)


server/go/README.md
```diff
@@ -54,6 +54,10 @@ Asherah Options:
                                           [$ASHERAH_CONNECTION_STRING]
       --enable-region-suffix              Configure the metastore to use regional suffixes (only supported by
                                           --metastore=dynamodb) [$ASHERAH_ENABLE_REGION_SUFFIX]
+      --dynamodb-endpoint=                An optional endpoint URL (hostname only or fully qualified URI) (only
```
sushantmimani, line 57: can simply call this `endpoint`

server/go/README.md
```diff
@@ -54,6 +54,10 @@ Asherah Options:
...
+                                          supported by --metastore=dynamodb) [$ASHERAH_DYNAMODB_ENDPOINT]
+      --dynamodb-region=                  The AWS region for DynamoDB requests (defaults to globally configured region)
```
sushantmimani, line 59: similarly, this can just be `region`

go/appencryption/pkg/persistence/dynamodb_test.go
```diff
@@ -339,10 +340,23 @@ func (suite *DynamoDBSuite) TestDynamoDBMetastore_Store_WithFailureShouldReturnE
 
 func (suite *DynamoDBSuite) TestDynamoDBMetastore_WithDynamoDBRegionSuffix() {
 	// keyPrefix should be empty unless WithDynamoDBRegionSuffix is used
-	assert.Empty(suite.T(), suite.dynamodbMetastore.keySuffix)
+	assert.Empty(suite.T(), suite.dynamodbMetastore.GetSuffix())
```
sushantmimani, line 343: can we rename this to be called `GetKeySuffix`

go/appencryption/pkg/persistence/dynamodb_test.go
```diff
@@ -339,10 +340,23 @@ func (suite *DynamoDBSuite) TestDynamoDBMetastore_Store_WithFailureShouldReturnE
 
 func (suite *DynamoDBSuite) TestDynamoDBMetastore_WithDynamoDBRegionSuffix() {
```
sushantmimani, line 341: unrelated to this PR but would it be better if we rename `WithDynamoDBRegionSuffix` to `WithKeySuffix`. Since this is in the `dynamodb.go` file, the DynamoDB in the name seems redundant, and since we call the variable keySuffix, `WithKeySuffix` seems to be more intuitive?

go/appencryption/pkg/persistence/dynamodb_test.go
```diff
@@ -339,10 +340,23 @@ func (suite *DynamoDBSuite) TestDynamoDBMetastore_Store_WithFailureShouldReturnE
...
-	assert.Equal(suite.T(), *suite.sess.Config.Region, suite.prefixedDynamodbMetastore.keySuffix)
+	assert.Equal(suite.T(), *suite.sess.Config.Region, suite.prefixedDynamodbMetastore.GetSuffix())
+}
+
+func (suite *DynamoDBSuite) TestDynamoDBMetastore_WithTableName() {
+	table := "DummyTable"
+	db := NewDynamoDBMetastore(suite.sess, WithTableName(table))
+
+	assert.Equal(suite.T(), table, db.tableName)
+}
```
sushantmimani, line 354: would be good to test that load/loadLatest works correctly in this test

go/appencryption/pkg/persistence/dynamodb_test.go
```diff
@@ -172,7 +173,7 @@ func (suite *DynamoDBSuite) SetupTest() {
 	}
 	suite.putItemInDynamoDB(getDynamoDBItem(en, suite.instant))
 
-	suite.dynamodbMetastore = NewDynamoDBMetastore(suite.sess)
+	suite.dynamodbMetastore = NewDynamoDBMetastore(suite.sess, WithTableName(tableName))
```
sushantmimani, line 176: should we keep using the defaults (table name in this case) for test and just add additional tests for new functional options?

server/go/pkg/server/options.go
```diff
@@ -15,6 +15,8 @@ type Options struct {
 	Metastore          string        `long:"metastore" choice:"rdbms" choice:"dynamodb" choice:"memory" required:"yes" description:"Determines the type of metastore to use for persisting keys" env:"ASHERAH_METASTORE_MODE"`
 	ConnectionString   string        `long:"conn" description:"The database connection string (required if --metastore=rdbms)" env:"ASHERAH_CONNECTION_STRING"`
 	EnableRegionSuffix bool          `long:"enable-region-suffix" description:"Configure the metastore to use regional suffixes (only supported by --metastore=dynamodb)" env:"ASHERAH_ENABLE_REGION_SUFFIX"`
+	DynamoDBEndpoint   string        `long:"dynamodb-endpoint" description:"An optional endpoint URL (hostname only or fully qualified URI) (only supported by --metastore=dynamodb)" env:"ASHERAH_DYNAMODB_ENDPOINT"`
+	DynamoDBRegion     string        `long:"dynamodb-region" description:"The AWS region for DynamoDB requests (defaults to globally configured region) (only supported by --metastore=dynamodb)" env:"ASHERAH_DYNAMODB_REGION"`
```
sushantmimani, line 19: should this not use the AWS_REGION env variable?

go/appencryption/pkg/persistence/dynamodb_test.go
aka-bo: I'd prefer to keep this as is. This is the `dynamodb.go` file, sure, but the file is part of the `persistence` package, so it's not really redundant. Renaming would mean users would reference the function as `persistence.WithKeySuffix(true)` which is ambiguous compared to the current `persistence.WithDynamoDBRegionSuffix(true)`.

go/appencryption/pkg/persistence/dynamodb_test.go
aka-bo: Using a custom table name for the tests retains the current integration-level testing coverage we had before and we assert the custom table name (and table name default) works via other targeted unit tests below, so all bases should be covered.

Ideally unit tests wouldn't require external dependencies at all (such as a local dynamodb instance), but considering the current state of testing here I figured this approach would be a decent compromise.

go/appencryption/pkg/persistence/dynamodb_test.go
aka-bo: shouldn't be needed (see comment re use of custom table name for tests)

server/go/pkg/server/options.go
aka-bo: Nope, the intent of this feature is to provide a means for the end-user to configure a region for DynamoDB only. AWS_REGION would potentially affect the region for ALL AWS services.

PR #219, opened 2020-06-08
Java: Rename method and variable

sushantmimani:
- Rename withRegionSuffix method to withKeySuffix
- Rename regionSuffix variable to keySuffix


nikoo28:
Making these variable name changes as they feel more intuitive and consistent with all SDK versions

PR #225, opened 2020-06-10
[Go] Use latest AE in the server and ref-app

aka-bo:
server:
- add --dynamodb-table-name flag

reference app: 
- new flags: --dynamodb-endpoint, --dynamodb-region, --dynamodb-table-name
- updated docs

### DIFF
--- a/samples/go/referenceapp/README.md
+++ b/samples/go/referenceapp/README.md
@@ -92,6 +92,10 @@ To use the DynamoDB Metastore included with Asherah, the following table should 
    ReadCapacityUnits=1,WriteCapacityUnits=1
 ```
 
+**NOTE**: To use a custom name for your DynamoDB table, replace `EncryptionKey` in the above command with your preferred
+table name. You will also need to configure the Reference App to use your custom table via the `--dynamodb-table-name`
+flag (example below).
+
 #### Global Tables
 
 To use Global Tables, the above table needs to be created with few modifications.
@@ -106,7 +110,7 @@ local DynamoDB endpoint.
 --metastore DYNAMODB \
 --enable-region-suffix \
 --dynamodb-endpoint http://localhost:8000 \
---dynamodb-table-name MyGlobalTable \
+--dynamodb-table-name MyCustomTableName \
 --payload-to-encrypt "some super secret value"
  ```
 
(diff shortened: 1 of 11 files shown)


samples/go/referenceapp/README.md
```diff
@@ -91,5 +91,19 @@ To use the DynamoDB Metastore included with Asherah, the following table should
...
+
+More details about how to create a Global Table can be found in the [AWS Developer Guide](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/globaltables.tutorial.html)
+
+Example run using DynamoDB metastore with regional suffixes enabled (for Global Tables) and static KMS
+
+```console
+[user@machine referenceapp]$ go run . \
+--metastore DYNAMODB \
+--enable-region-suffix \
+--payload-to-encrypt "some super secret value"
```
nikoo28, line 106: since we have had the feature request to enable endpoint configuration, it'd be good to show here how it can be passed. Similarly, can also add the parameter for a table name

samples/go/referenceapp/README.md
```diff
@@ -91,5 +91,19 @@ To use the DynamoDB Metastore included with Asherah, the following table should
    ReadCapacityUnits=1,WriteCapacityUnits=1
 ```
```
nikoo28, line 92: We can mention in the above example that in case of DynamoDb, you can use a custom table name on L:83

server/go/pkg/server/server_test.go
```diff
@@ -124,6 +124,7 @@ func Test_Streamer_NewHandler(t *testing.T) {
 		{KMS: "aws", Metastore: "dynamodb"},
 		{KMS: "aws", Metastore: "dynamodb", DynamoDBRegion: "us-east-1"},
 		{KMS: "aws", Metastore: "dynamodb", DynamoDBEndpoint: "http://localhost:8000"},
+		{KMS: "aws", Metastore: "dynamodb", DynamoDBTableName: "CustomTable"},
```
nikoo28, line 127:
```suggestion
		{KMS: "aws", Metastore: "dynamodb", DynamoDBTableName: "CustomTableName"},
```

PR #228, opened 2020-06-13
[C#] : Update reference app and documentation

nikoo28:
- updated documentation for global tables in the SDK
- updated documentation for configuring the reference app

### DIFF
--- a/csharp/AppEncryption/README.md
+++ b/csharp/AppEncryption/README.md
@@ -77,13 +77,33 @@ IMetastore<JObject> adoMetastore = AdoMetastoreImpl.NewBuilder(dbProviderFactory
 ```
 
 #### DynamoDB Metastore
+For simplicity, the DynamoDB implementation uses the builder pattern to enable configuration changes.
+
+To obtain an instance of the builder, use the static factory method `NewBuilder`. 
+```c#
+DynamoDbMetastoreImpl.NewBuilder();
+```
+Once you have a builder, you can either use the `WithXXX` setter methods to configure the metastore properties or simply
+build the metastore by calling the `Build` method.
+
+ - **WithKeySuffix**: Specifies whether key suffix should be enabled for DynamoDB. **This is required to enable Global
+ Tables**.
+ - **WithTableName**: Specifies the name of the DynamoDb table.
+ - **WithRegion**: Specifies the region for the AWS DynamoDb client.
+ - **WithEndPointConfiguration**: Adds an EndPoint configuration to the AWS DynamoDb client.
+
+Below is an example of a DynamoDB metastore that use a Global Table named `TestTable`
 
 ```c#
 // Setup region via global default or via other AWS .NET SDK mechanisms
 AWSConfigs.AWSRegion = "us-west-2";
 
 // Build the DynamoDB Metastore.
+Metastore dynamoDbMetastore = DynamoDbMetastoreImpl.NewBuilder()
 IMetastore<JObject> dynamoDbMetastore = DynamoDbMetastoreImpl.NewBuilder().Build();
+      .WithKeySuffix("us-west-2")
+      .WithTableName("TestTable")
+      .Build();
 ```
 
 #### In-memory Metastore (FOR TESTING ONLY)
--- a/csharp/AppEncryption/README.md
+++ b/csharp/AppEncryption/README.md
@@ -92,7 +92,7 @@ build the metastore by calling the `Build` method.
  - **WithRegion**: Specifies the region for the AWS DynamoDb client.
  - **WithEndPointConfiguration**: Adds an EndPoint configuration to the AWS DynamoDb client.
 
-Below is an example of a DynamoDB metastore that use a Global Table named `TestTable`
+Below is an example of a DynamoDB metastore that uses a Global Table named `TestTable`
 
 ```c#
 // Setup region via global default or via other AWS .NET SDK mechanisms
--- a/samples/csharp/ReferenceApp/README.md
+++ b/samples/csharp/ReferenceApp/README.md
@@ -128,17 +128,15 @@ Configuration options are provided via command-line arguments. Supported options
 follows:
 
 ```console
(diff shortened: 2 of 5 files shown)


samples/csharp/ReferenceApp/ReferenceApp/ReferenceApp.cs
```diff
@@ -81,7 +81,53 @@ private static async void App(Options options)
...
+                    if (string.IsNullOrEmpty(options.DynamodbEndpoint) ||
+                        string.IsNullOrEmpty(options.DynamodbSigningRegion))
+                    {
+                        logger.LogError("One or more parameter(s) for endpoint configuration missing.");
+                        Console.WriteLine(HelpText.AutoBuild(cmdOptions, null, null));
+                        return;
+                    }
+
+                    builder.WithEndPointConfiguration(options.DynamodbEndpoint, options.DynamodbSigningRegion);
+                }
```
aka-bo, line 118: This validation is a bit cumbersome and potentially confusing. If the `DynamodbRegion` and `DynamodbSigningRegion` flags are mutually exclusive yet both used to provide a region name for different config options, can we just combine them into a single flag, i.e., `DynamodbRegion`?

Then we could simplify with something like this:
```suggestion
                if (!string.IsNullOrEmpty(options.DynamodbEndpoint))
                {
                    if (string.IsNullOrEmpty(options.DynamodbRegion))
                    {
                        // TODO: check if region can be determined from provided endpoint
                        logger.LogError("DynamoDb region is required when providing a DynamoDb endpoint.");
                        Console.WriteLine(HelpText.AutoBuild(cmdOptions, null, null));
                        return;
                    }

                    builder.WithEndPointConfiguration(options.DynamodbEndpoint, options.DynamodbRegion);
                }
                else if (!string.IsNullOrEmpty(options.DynamodbRegion))
                {
                    builder.WithRegion(options.DynamodbRegion);
                }
```

csharp/AppEncryption/README.md
```diff
@@ -77,13 +77,33 @@ IMetastore<JObject> adoMetastore = AdoMetastoreImpl.NewBuilder(dbProviderFactory
...
+Once you have a builder, you can either use the `WithXXX` setter methods to configure the metastore properties or simply
+build the metastore by calling the `Build` method.
+
+ - **WithKeySuffix**: Specifies whether key suffix should be enabled for DynamoDB. **This is required to enable Global
+ Tables**.
+ - **WithTableName**: Specifies the name of the DynamoDb table.
+ - **WithRegion**: Specifies the region for the AWS DynamoDb client.
+ - **WithEndPointConfiguration**: Adds an EndPoint configuration to the AWS DynamoDb client.
+
+Below is an example of a DynamoDB metastore that use a Global Table named `TestTable`
```
sushantmimani, line 95:
```suggestion
Below is an example of a DynamoDB metastore that uses a Global Table named `TestTable`
```

samples/csharp/ReferenceApp/README.md
```diff
@@ -97,6 +103,60 @@ To use the DynamoDB Metastore included with the App Encryption library, the foll
...
+
+  -e, --dynamodb-endpoint        The DynamoDb service endpoint (only supported by DYNAMODB)
+
+  -r, --dynamodb-region          The AWS region for DynamoDB requests (only supported by DYNAMODB)
+
+  -s, --dynamodb-signing-region  The DynamoDb service endpoint (only supported by DYNAMODB)
+
+  -t, --dynamodb-table-name      The table name for DynamoDb (only supported by DYNAMODB)
+
+  -x, --key-suffix               Configure the metastore to use key suffixes (only supported by DYNAMODB)
```
sushantmimani, line 141: can we use `-s` instead?

samples/csharp/ReferenceApp/ReferenceApp/ReferenceApp.cs
```diff
@@ -81,7 +81,53 @@ private static async void App(Options options)
...
+
+                if (!string.IsNullOrEmpty(options.DynamodbTableName))
+                {
+                    builder.WithTableName(options.DynamodbTableName);
+                }
+
+                if (!string.IsNullOrEmpty(options.KeySuffix))
+                {
+                    builder.WithKeySuffix(options.KeySuffix);
+                }
```
sushantmimani, line 128: discussed offline, this needs an additional check for  empty strings

samples/csharp/ReferenceApp/README.md
nikoo28: this was done when I added `s` for the `singing-region`, and hence `x` seemed to be closest to `suffix`...
anyway, we can keep `s` again now. made the change.

samples/csharp/ReferenceApp/ReferenceApp/ReferenceApp.cs
```diff
@@ -81,7 +81,43 @@ private static async void App(Options options)
...
+                    builder.WithEndPointConfiguration(options.DynamodbEndpoint, options.DynamodbRegion);
+                }
+                else if (!string.IsNullOrEmpty(options.DynamodbRegion))
+                {
+                    builder.WithRegion(options.DynamodbRegion);
+                }
+
+                if (!string.IsNullOrEmpty(options.DynamodbTableName))
+                {
+                    builder.WithTableName(options.DynamodbTableName);
```
sushantmimani, line 105: needs a similar 2 level check like key suffix

samples/csharp/ReferenceApp/ReferenceApp/ReferenceApp.csproj
```diff
@@ -18,7 +18,7 @@
         <!-- End of Properties related to NuGet packaging: -->
     </PropertyGroup>
     <ItemGroup>
-        <PackageReference Include="GoDaddy.Asherah.AppEncryption" Version="0.1.1" />
+        <PackageReference Include="GoDaddy.Asherah.AppEncryption" Version="0.1.3" />
```
sushantmimani, line 21: should now use 0.1.4

PR #234, opened 2020-06-16
[Java]Move scripts to a central location and symlink everywhere

sushantmimani:


nikoo28:
Can probably move the scripts for `Go` and `C#` as well with this change itself.

PR #238, opened 2020-06-17
[Go] Move scripts to root level build directory and symlink everywhere

sushantmimani:
- remove redundant scripts
- bump up go versions to `alpha`


java/secure-memory/pom.xml
```diff
@@ -4,7 +4,7 @@
 
   <groupId>com.godaddy.asherah</groupId>
   <artifactId>securememory</artifactId>
-  <version>0.1.0</version>
+  <version>0.1.1-alpha</version>
```
aka-bo, line 7: Is the version bump necessary? `java/securememory/v0.1.0` has already been released so leaving the version as-is would skip the deploy phase, which is effectively the same as bumping the version to `0.1.1-alpha`, right? If the intent is to make the "no-op deploy" explicit, I think that's ok. It just gets a bit confusing when we're bumping the version for Java implementations as part of a go-centric PR.

sushantmimani: not strictly necessary for this PR. The intent is to make the "no-op deploy" explicit, but i can make that change whenever the next batch of minor Java changes happen.

PR #241, opened 2020-06-19
[C#] Update AwsKeyManagementServiceImplTest

sushantmimani:
- Newtonsoft.Json has an [issue](https://github.com/JamesNK/Newtonsoft.Json/issues/1437) where DeepEquals fails when order of array elements is different. Adding a custom checker to verify if the test should really fail or if it's just failing due to the aforementioned issue.

### DIFF
--- a/csharp/AppEncryption/AppEncryption.Tests/AppEncryption/Kms/AwsKeyManagementServiceImplTest.cs
+++ b/csharp/AppEncryption/AppEncryption.Tests/AppEncryption/Kms/AwsKeyManagementServiceImplTest.cs
@@ -4,6 +4,7 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Collections.Specialized;
 using System.IO;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Amazon.KeyManagementService;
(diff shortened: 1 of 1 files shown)


csharp/AppEncryption/AppEncryption.Tests/AppEncryption/Kms/AwsKeyManagementServiceImplTest.cs
```diff
@@ -363,8 +364,6 @@ private void TestEncryptKeySuccessful()
             {
                 { EncryptedKey, Convert.ToBase64String(encryptedKey) },
                 {
-                    // For some reason we have to use ConcurrentBag here. Likely due to internal ConcurrentBag list
-                    // structure failing to compare against regular List?
                     KmsKeksKey, new ConcurrentBag<object>
```
aka-bo, line 367: Can use a `List<object>` here and drop the `.ToList()` on L376

sushantmimani: Potentially, yes. The reason for doing it this was to keep the process of the envelope creation the same as the KMS implementation. https://github.com/godaddy/asherah/blob/e76a2c0cdf1cd9edee5ff64132130a78d127cbec/csharp/AppEncryption/AppEncryption/Kms/AwsKeyManagementServiceImpl.cs#L149

Also, I think  that using a List directly caused the test to fail intermittently in the past, as mentioned in the now deleted comment. Since the test now does exactly what we do in the code, the comment felt unnecessary.

sushantmimani: Discussed offline. Using a List<object> for the test

PR #254, opened 2020-07-08
RDBMS connection pool

jeffchannell:
The Golang sidecar does not seem to ever relinquish its database connections. After running a test application a number of times, there were database connections lingering in MySQL well after the test runs were complete.

This patch adds an option to use the same database connection for every session.

Without the patch, or without the pool option being enabled, this is the resulting process list in MySQL after running the test application a number of times:

```
+-----+----------+-----------------+---------+---------+------+----------+------------------+
| Id  | User     | Host            | db      | Command | Time | State    | Info             |
+-----+----------+-----------------+---------+---------+------+----------+------------------+
| 165 | testuser | 127.0.0.1:54726 | asherah | Sleep   |   99 |          | NULL             |
| 237 | testuser | 127.0.0.1:54898 | asherah | Sleep   |   81 |          | NULL             |
| 242 | testuser | 127.0.0.1:54914 | asherah | Sleep   |   80 |          | NULL             |
| 312 | testuser | 127.0.0.1:55076 | asherah | Sleep   |   62 |          | NULL             |
| 319 | testuser | 127.0.0.1:55094 | asherah | Sleep   |   60 |          | NULL             |
| 389 | testuser | 127.0.0.1:55256 | asherah | Sleep   |   43 |          | NULL             |
| 396 | testuser | 127.0.0.1:55274 | asherah | Sleep   |   41 |          | NULL             |
| 466 | testuser | 127.0.0.1:55436 | asherah | Sleep   |   23 |          | NULL             |
| 473 | testuser | 127.0.0.1:55454 | asherah | Sleep   |   21 |          | NULL             |
| 488 | testuser | localhost       | NULL    | Query   |    0 | starting | show processlist |
+-----+----------+-----------------+---------+---------+------+----------+------------------+
```

After enabling the connection pool option provided by this patch, the resulting process list contains only one connection:

```
+-----+----------+-----------------+---------+---------+------+----------+------------------+
| Id  | User     | Host            | db      | Command | Time | State    | Info             |
+-----+----------+-----------------+---------+---------+------+----------+------------------+
| 127 | testuser | 127.0.0.1:55844 | asherah | Sleep   |    8 |          | NULL             |
| 361 | testuser | localhost       | NULL    | Query   |    0 | starting | show processlist |
+-----+----------+-----------------+---------+---------+------+----------+------------------+
```

There is also the possibility of rejecting this patch in favor of adding a method to explicitly end a session which would close the database connection, leaving it to the client to close the session once it is finished using it. I considered this approach but it seemed like a much bigger task.


server/go/pkg/server/database.go
```diff
@@ -6,6 +6,27 @@ import (
 	"github.com/go-sql-driver/mysql"
 )
 
+var (
+	dbconnection *sql.DB
+)
+
+func getMysql(connStr string, pool bool) (*sql.DB, error) {
```
aka-bo, line 13: can merge this function into `newMysql` if we eliminate the new option as recommended.

server/go/pkg/server/database.go
```diff
@@ -6,6 +6,27 @@ import (
...
+	if !pool {
+		return newMysql(connStr)
+	}
+
+	if (*sql.DB)(nil) == dbconnection {
```
aka-bo, line 18: a vanilla `nil` check will suffice as long as we never store `(*sql.DB)(nil)` in the interface var. Should also add a basic unit test to cover the new logic (time to create a database_test.go perhaps?).

server/go/pkg/server/options.go
```diff
@@ -10,17 +10,18 @@ import (
 type Options struct {
 	ServiceName        string        `long:"service" required:"yes" description:"The name of this service" env:"ASHERAH_SERVICE_NAME"`
 	ProductID          string        `long:"product" required:"yes" description:"The name of the product that owns this service" env:"ASHERAH_PRODUCT_NAME"`
-	ExpireAfter        time.Duration `long:"expire-after" description:"The amount of time a key is considered valid" env:"ASHERAH_EXPIRE_AFTER"`
```
aka-bo, line 13: these changes can be reverted

server/go/pkg/server/server.go
```diff
@@ -99,7 +99,7 @@ func NewMetastore(opts *Options) appencryption.Metastore {
 	switch opts.Metastore {
 	case "rdbms":
 		// TODO: support other databases
-		db, err := newMysql(opts.ConnectionString)
```
aka-bo, line 102: can also revert this

Issue #247, opened 2020-06-29
[JAVA] Add java integration tests in the CICD pipeline.


nikoo28:
Currently the integration tests for java implementation of SDK are written as a separate test-app and they do not run in the CICD pipeline.
Find a way to integrate these tests in the CICD pipeline. One way could be to add them in the project SDK itself.

PR #255, opened 2020-07-08
Integration tests migration

nikoo28:
Moved integration tests from `test-app` to the library itself.
Additional task to remove the `test-app` job and its contents will be done at a later stage.

Additional integration tests to verify cross partition encryption/decryption would be written once this gets merged in.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -194,6 +194,7 @@ jobs:
       - run:
           name: Integration Tests
           command: |
+            source ./.circleci/set_env_variables.sh
             cd java/app-encryption
             ./scripts/integration_test.sh
       - store_test_results:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -191,6 +191,11 @@ jobs:
             source ./.circleci/set_env_variables.sh
             cd java/app-encryption
             ./scripts/test.sh
+      - run:
+          name: Integration Tests
+          command: |
+            cd java/app-encryption
+            ./scripts/integration_test.sh
       - store_test_results:
           path: java/app-encryption/target/surefire-reports
       - codecov/upload:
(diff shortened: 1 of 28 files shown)


.circleci/config.yml
```diff
@@ -191,6 +191,12 @@ jobs:
             source ./.circleci/set_env_variables.sh
             cd java/app-encryption
             ./scripts/test.sh
+      - run:
```
sushantmimani, line 194: Think it should be ok to remove the test-app step from circleci in this PR, even if the actual directory is removed later

java/app-encryption/src/it/java/com/godaddy/asherah/ConfigReader.java
```diff
@@ -0,0 +1,30 @@
+package com.godaddy.asherah;
+
+public class ConfigReader {
```
sushantmimani, line 3: Rename this class to something like `TestConfiguration` as we really aren't reading anything here

java/app-encryption/src/it/java/com/godaddy/asherah/ConfigReader.java
```diff
@@ -0,0 +1,30 @@
...
+
+  private String metastoreType;
+  private String metastoreJdbcUrl;
```
sushantmimani, line 6: Add params for username, password, connectionpool size (like the existing test-app) for more flexibility.

java/app-encryption/src/it/resources/config.yaml
```diff
@@ -0,0 +1,5 @@
+metastoreType: "memory"
+metastoreJdbcUrl: ""
```
sushantmimani, line 2: add params for username, password, connectionPoolSize for more flexible configuration

nikoo28:
This closes https://github.com/godaddy/asherah/issues/247

Issue #259, opened 2020-07-15
Dependency on StyleCop creates masses of build warnings


lyricnz:
When adding Asherah AppEncryption to our project via NuGet, it drags in Asherah.Crypto which in turn has a dependency on StyleCop.Analyzers.  This creates hundreds of warnings in our project builds, clogging logs and making the IDE very messy / hard-to-ues.

Is there some way in which we can disable StyleCop completely?  We don't it running on our projects.

I tried to find ways to turn this off myself, but was unsuccessful.

nikoo28:
looks like there is a way to have some libraries for development use only. this is being worked upon.

lyricnz:
Version 0.1.5 appears to resolve this. Thanks!

PR #261, opened 2020-07-15
[C#] Logging: Convert StyleCop to a development dependency

sushantmimani:
Step 1 of 3 towards resolving https://github.com/godaddy/asherah/issues/259

### DIFF
--- a/csharp/Logging/Directory.Build.props
+++ b/csharp/Logging/Directory.Build.props
@@ -1,5 +1,5 @@
 <Project>
   <PropertyGroup>
-    <Version>0.1.3-alpha</Version>
+    <Version>0.1.3</Version>
   </PropertyGroup>
 </Project>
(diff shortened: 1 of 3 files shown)


csharp/Logging/Directory.Build.props
```diff
@@ -1,5 +1,5 @@
 <Project>
   <PropertyGroup>
-    <Version>0.1.3-alpha</Version>
+    <Version>0.1.3</Version>
```
nikoo28, line 3: this got removed unintentionally?

nikoo28:
Can also probably add some description to the PR stating why this change is being made, or refer to the issue.

csharp/Logging/Directory.Build.props
sushantmimani: yes, because this will be release

PR #269, opened 2020-07-16
Fix JavaDoc issues for AE v0.1.3 release

sushantmimani:
- Moved the `maven-javadoc-plugin` out of the release profile block. This will ensure that JavaDocs are generated and validated for every build.


nikoo28:
looks good now...

PR #272, opened 2020-07-20
[go] session caching functionality

aka-bo:
Adds initial support for shared session cache functionality to the go appencryption implementation.

* Supports two underlying cache packages:
  * [goburrow/cache (aka Mango)](https://github.com/goburrow/cache) - A stable, mature cache package. This is the default.
  * [dgraph-io/ristretto](https://github.com/dgraph-io/ristretto) - A promising new high-perf cache library under active development.
* Updated performance test app (go/appencryption/cmd/example) to support session caching
* doc updates
* go/appencryption v0.1.3
* Marked TDD workflow improvements - slow tests have been instrumented using `testing.Short()` and can now be skipped using `go test -short`


go/appencryption/session_cache_test.go
```diff
@@ -0,0 +1,376 @@
...
+		defer cache.Close()
+
+		require.NotNil(t, cache)
+	})
+}
+
+func withEachEngine(t *testing.T, testFunc func(*testing.T, *appencryption.CryptoPolicy)) {
+	var engines = [...]string{
+		"mango",
+		"ristretto",
```
nikoo28, line 108: those constants can then be used here

go/appencryption/session_cache_test.go
```diff
@@ -0,0 +1,376 @@
...
+
+			time.Sleep(time.Millisecond * 200)
+
+			// shared sessions arent' actually closed until evicted from the cache
+			assert.False(t, b.IsClosed(s1))
+		}
+	})
+}
+
+func TestSharedSessionNotClosedWhenInUse(t *testing.T) {
```
nikoo28, line 350: can we rename this similar to `TestSessionCacheGetSessionWhileStillUsedAndNotExpiredShouldNotEvict`

go/appencryption/session_cache_test.go
```diff
@@ -0,0 +1,376 @@
...
+
+		assert.Eventually(t, func() bool {
+			return b.IsClosed(s1)
+		}, time.Second*10, time.Millisecond*100)
+
+		assert.False(t, b.IsClosed(s2))
+	})
+}
+
+func TestSharedSessionClose(t *testing.T) {
```
nikoo28, line 323: can we rename this similar to `TestSessionCacheGetSessionAfterUseAndExpiredShouldEvict`

aka-bo:
> * just like metastore is an interface and dynmaodb and inmemory are different kind of implementations.
>   Can we use the session cache in a similar way? Making `ristretto` and `mango` as different implementations?

The underlying cache implementation is not intended to be pluggable and I fully expect one of these to be phased out in the not-too-distant future. If that assumption proves to be false and we find value in supporting multiple implementations long term then we can revisit, for now though I'd like to keep the public API changes to a minimum.

> * pending multi-threaded tests for session cache.

Can you elaborate on this point? Are you requesting additional integration and/or functional tests?

go/appencryption/policy.go
```diff
@@ -26,6 +29,16 @@ type CryptoPolicy struct {
 	CacheIntermediateKeys bool
 	// CacheSystemKeys determines whether System Keys will be cached.
 	CacheSystemKeys bool
+	// CacheSessions determines whether sessions will be cached.
+	CacheSessions bool
+	// SessionCacheSize controls the maximum size of the cache if session caching is enabled.
+	SessionCacheSize int
```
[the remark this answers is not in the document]
aka-bo: already done as part of 7ded942... guessing github was showing you an older version

go/appencryption/README.md
```diff
@@ -183,6 +184,23 @@ cryptoPolicy := appencryption.NewCryptoPolicy(
...
+
+To enable session caching, simply use the `WithSessionCache` option when creating the crypto policy.
+
+```go
+cryptoPolicy := appencryption.NewCryptoPolicy(
+    appencryption.WithExpireAfterDuration(24 * time.Hour),
+    appencryption.WithRevokeCheckInterval(30 * time.Minute),
+    appencryption.WithSessionCache(),
+    appencryption.SessionCacheMaxSize(200),
+    appencryption.WithSessionCacheTTL(5 * time.Minute),
```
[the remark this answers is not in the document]
aka-bo: ah, good catch

go/appencryption/session_cache_test.go
aka-bo: actually meant to remove this test as it's functionally equivalent to `TestSharedSessionClose`. will remove

go/appencryption/session_cache_test.go
aka-bo: This test is meant to assert that calling closed on a cached session does not result in a close of the underlying encryption object. will rename accordingly

go/appencryption/session_cache.go
```diff
@@ -0,0 +1,168 @@
...
+			SessionInjectEncryption(s, wrapped)
+		}
+
+		return s, nil
+	}
+
+	switch eng := policy.SessionCacheEngine; eng {
+	case "", "default", "mango":
+		return newMangoCache(wrapper, policy)
+	case "ristretto":
```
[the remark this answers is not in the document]
aka-bo: This is the only occurrence of these string literals in the code (apart from tests and comments). I'd prefer we avoid adding a package level constant until we need to make use of these values elsewhere (at which point we'd likely look at other refactoring away that need entirely)

go/appencryption/session_cache_test.go
```diff
@@ -0,0 +1,379 @@
...
+
+		defer cache.Close()
+
+		val, err := cache.Get("some-id")
+		require.NoError(t, err)
+		assert.Same(t, session, val)
+	})
+}
+
+func TestSessionCacheGetDoesNotUseLoaderOnHit(t *testing.T) {
```
sushantmimani, line 155: This test verifies that if we find a session in the cache, then the cache loader isn't called. Should/can we also have a check that verifies that when we find a session in the cache, then there are no metastore calls?

go/appencryption/session_cache.go
```diff
@@ -0,0 +1,168 @@
...
+			mango.WithMaximumSize(policy.SessionCacheMaxSize),
+			mango.WithExpireAfterAccess(policy.SessionCacheDuration),
+			mango.WithRemovalListener(mangoRemovalListener),
+		),
+	}
+}
+
+// sharedEncryption is used to track the number of concurrent users to ensure sessions remain
+// cached while in use.
+type sharedEncryption struct {
```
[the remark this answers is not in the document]
aka-bo: discussed offline. will leave this as-is and consider renaming the java and c# to eliminate confusion between envelope encryption implementations and sessions

go/appencryption/session_cache_ristretto.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/dgraph-io/ristretto"
+)
+
+// ristrettoCache is a SessionCache implementation based on dgraph-io's
+// Ristretto cache library.
+type ristrettoCache struct {
+	inner   *ristretto.Cache
+	loader  sessionLoaderFunc
+	ttl     time.Duration
+	maxSize int64
```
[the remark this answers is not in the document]
aka-bo: discussed offline. will leave as-is as `policy` is unavailable outside of the constructor

sushantmimani:
Discussed that multi-threaded(benchmark) tests will be added as a separate PR, but before making a new release. Everything else looks good.

PR #274, opened 2020-07-23
[C#]Use latest version of AE

sushantmimani:
- Update reference app
- Update documentation

### DIFF
--- a/samples/csharp/ReferenceApp/README.md
+++ b/samples/csharp/ReferenceApp/README.md
@@ -117,7 +117,7 @@ local DynamoDB endpoint.
 ```console
 [user@machine ReferenceApp]$ dotnet bin/Debug/netcoreapp2.0/ReferenceApp.dll \
   --metastore-type DYNAMODB \
-  --key-suffix us-west-2 \
+  --key-suffix \
   --dynamodb-endpoint http://localhost:8000 \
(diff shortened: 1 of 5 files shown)


samples/csharp/ReferenceApp/ReferenceApp/Options.cs
```diff
@@ -19,7 +19,7 @@ public class Options
         public string DynamodbTableName { get; set; }
 
         [Option('s', "key-suffix", Required = false, HelpText = "Configure the metastore to use key suffixes (only supported by DYNAMODB)")]
-        public string KeySuffix { get; set; }
+        public bool KeySuffix { get; set; }
```
nikoo28, line 22: this should change to `EnableKeySuffix` since it is now a bool value.

samples/csharp/ReferenceApp/README.md
```diff
@@ -117,7 +117,7 @@ local DynamoDB endpoint.
 ```console
 [user@machine ReferenceApp]$ dotnet bin/Debug/netcoreapp2.0/ReferenceApp.dll \
   --metastore-type DYNAMODB \
-  --key-suffix us-west-2 \
+  --key-suffix \
```
nikoo28, line 120: this should just be `--enable-key-suffix`

PR #283, opened 2020-09-08
[go] v0.1.3 updates for server and referenceapp

aka-bo:
Server:
- upgrade appencryption to v0.1.3
- added new configuration options to support shared session caching
- all configuration options and their configured values are now logged on startup
- fixed a bug that resulted in the creation of a new session factory on every GetSession request

ReferenceApp:
- upgrade appencryption to v0.1.3


sushantmimani (approved):
LGTM 🦅

PR #287, opened 2020-09-16
Integrate WhiteSource scans as a scheduled workflow

sushantmimani:
Add a new workflow that runs WSS scans at 11:00 every Monday.
The scan will run only on the master branch on the godaddy/asherah repo (not on forks)


aka-bo (approved):
LGTM

PR #297, opened 2020-09-28
secure-memory: JDK 11 support (build env only), upgraded dependencies

aka-bo:
Note: compilation target continues to be JDK-8


sushantmimani (approved):
LGTM 🦅

PR #300, opened 2020-09-29
csharp OpenSSL phase 2

jgowdy:
C# OpenSSL phase 2 is largely about modifying AppEncryption to support IConfiguration based configuration instead of or complementary to the existing builder pattern, which also enables passing IConfiguration down to the SecureMemory layer for selection of the OpenSSL secure heap.

See also this PR to SecureMemory which AppEncryption is now updated to: https://github.com/godaddy/asherah/pull/324

Phase 3 should be the final PR which will move to using native OpenSSL crypto facilties, which allows us to never have encryption keys in a decrypted state in managed memory.


jgowdy:
https://github.com/godaddy/asherah/pull/318#issuecomment-705855701

jgowdy:
https://github.com/godaddy/asherah/blob/baa290e41afad69b5a98183003ccce53b7c37363/csharp/AppEncryption/README.md#configuration

jgowdy:
Code coverage is as high as possible, uncovered lines are largely fatal and difficult to test edge cases.

jgowdy:
I'm changing direction on this

lyricnz:
@jgowdy whatever happened with this, man?  "changed direction" to what?   we are still having issues with mlock exhaustion (on AWS Lambda environments), and were hoping for openssl-heap based solution to avoid excessive mlock.

PR #304, opened 2020-10-01
Add creation stack trace debug mode for ProtectedMemorySecret

jgowdy:
Since failure to Dispose of a ProtectedMemorySecret is now a fatal error, it becomes important to be able to figure out which secret usage isn't being disposed of.  Since the fatal error happens in the finalizer thread, there's no good way to identify that.  With this additional bit of configuration for debugging secrets enabled, ProtectedMemorySecret will save the stack trace of the constructor call and include that in the exception message if it hits the finalizer.


aka-bo (approved):
LGTM

PR #306, opened 2020-10-01
Fix NullReferenceException in null config case

jgowdy:


csharp/SecureMemory/SecureMemory.Tests/ProtectedMemoryImpl/ProtectedMemorySecretTest.cs
```diff
@@ -44,6 +44,16 @@ public void Dispose()
...
 
+        [Fact]
+        private void TestNullConfiguration()
+        {
+            Debug.WriteLine("TestNullConfiguration");
+            protectedMemoryAllocatorMock.Setup(x => x.Alloc(It.IsAny<ulong>())).Returns(IntPtr.Zero);
+            using (var secret = new ProtectedMemorySecret(new byte[] { 0, 1 }, protectedMemoryAllocatorMock.Object, null))
+            {
+            }
+        }
```
aka-bo, line 55: Test currently fails. Using an actual allocator will fix it up

```suggestion
        [Theory]
        [ClassData(typeof(AllocatorGenerator))]
        private void TestNullConfiguration(IProtectedMemoryAllocator protectedMemoryAllocator)
        {
            Debug.WriteLine("TestNullConfiguration");
            using (var secret = new ProtectedMemorySecret(new byte[] { 0, 1 }, protectedMemoryAllocator, null))
            {
            }
        }
```

aka-bo (changes requested):
just need to fix up the test

aka-bo (approved):

PR #311, opened 2020-10-03
[C#] Update DynamoDbMetastoreImpl to throw detailed AWS exceptions

sushantmimani:
TryLoadTable just returns a true or a false based on whether it was able to load a table.
Due to this, we can run in NPE if a table was  not loaded and we try to run operations on it.
Replacing it with LoadTable because this throws an exception if the table isn't loaded.


csharp/AppEncryption/AppEncryption.Tests/AppEncryption/Persistence/DynamoDbMetastoreImplTest.cs
```diff
@@ -42,12 +43,18 @@ public class DynamoDbMetastoreImplTest : IClassFixture<DynamoDBContainerFixture>
...
+                ServiceURL = dynamoDbContainerFixture.ServiceUrl,
+                AuthenticationRegion = "us-west-2",
+            };
+            amazonDynamoDbClient = new AmazonDynamoDBClient(clientConfig);
+
+            CreateTableSchema(amazonDynamoDbClient, "EncryptionKey");
+
             dynamoDbMetastoreImpl = NewBuilder(Region)
-                .WithEndPointConfiguration(dynamoDbContainerFixture.ServiceUrl, "us-west-2")
+                .WithEndPointConfiguration("http://localhost:8000", Region)
```
aka-bo, line 56: Why not use `dynamoDbContainerFixture.ServiceUrl` here?

csharp/AppEncryption/AppEncryption.Tests/AppEncryption/Persistence/DynamoDbMetastoreImplTest.cs
```diff
@@ -276,6 +283,7 @@ private void TestPrimaryBuilderPath()
             // Hack to inject default region since we don't explicitly require one be specified as we do in KMS impl
             AWSConfigs.AWSRegion = "us-west-2";
             DynamoDbMetastoreImpl dbMetastoreImpl = NewBuilder(Region)
+                .WithEndPointConfiguration("http://localhost:" + DynamoDbPort, Region)
```
aka-bo, line 286: This test is now functionally equivalent to the test that follows. Do we need both?

csharp/AppEncryption/AppEncryption.Tests/AppEncryption/Persistence/DynamoDbMetastoreImplTest.cs
sushantmimani: fixed

csharp/AppEncryption/AppEncryption.Tests/AppEncryption/Persistence/DynamoDbMetastoreImplTest.cs
sushantmimani: reverted this test back to the original commented out version. Think it would be good to leave it in there so we can uncomment and verify that the primary builder path is not broken when we make any new changes to the Dynamo implementation.

aka-bo (approved):
LGTM

PR #323, opened 2020-10-14
Update nuget dependency resolver

sushantmimani:
Set `nuget.resolveDependencies` to true so that all dependencies are resolved before running the scan. This should take care of the libraries not getting updated correctly for the C# projects.


aka-bo (approved):
LGTM

PR #324, opened 2020-10-15
Fix the locking on ProtectedMemorySecret

jgowdy:
The locking on ProtectedMemorySecret wasn't done correctly, and it wasn't being caught well by the tests back when we weren't universally calling Dispose.  When we started calling Dispose more often, we would get phantom errors during or after Dispose calls when Interlocked.Exchange would yank the pointer away after the IntPtr.Zero / null check outside of any synchronized context.

In this new version, we use a ReaderWriterLockSlim.EnterReadLock for all read access to the secret, and we use ReaderWriterLockSlim.EnterWriteLock for Dispose.  This allows multiple concurrent readers of the same secret unlike the previous version, and ensures a synchronized context when disposing.

Also in this PR, we ensure that the managed byte[] and managed char[] used in WithSecretBytes and WithSecretUtf8Chars (which is misnamed) are both GC pinned while they're being accessed until they're wiped.  This keeps the garbage collector from relocating them and potentially leaking copies of the secrets.

This also introduces WithSecretIntPtr which allows for the unmanaged pointer to be used directly with other unmanaged calls like OpenSSL crypto calls in future work.

### DIFF
--- a/csharp/SecureMemory/SecureMemory.Tests/ProtectedMemoryImpl/Libc/LibcProtectedMemoryAllocatorTest.cs
+++ b/csharp/SecureMemory/SecureMemory.Tests/ProtectedMemoryImpl/Libc/LibcProtectedMemoryAllocatorTest.cs
@@ -130,6 +130,17 @@ namespace GoDaddy.Asherah.SecureMemory.Tests.ProtectedMemoryImpl.Libc
         }
 
         [SkippableFact]
+        private void TestFreeWithInvalidLengthShouldFail()
+        {
+            Skip.If(libc == null);
+
+            Debug.WriteLine("LibcProtectedMemoryAllocatorTest.TestCheckPointerWithRegularPointerShouldSucceed");
+
+            IntPtr fakePtr = IntPtr.Add(IntPtr.Zero, 1);
+            Assert.Throws<LibcOperationFailedException>(() => libcProtectedMemoryAllocator.Free(fakePtr, 0));
+        }
+
+        [SkippableFact]
         private void TestCheckPointerWithNullPointerShouldFail()
         {
             Skip.If(libc == null);
--- a/csharp/SecureMemory/SecureMemory.Tests/ProtectedMemoryImpl/Libc/LibcProtectedMemoryAllocatorTest.cs
+++ b/csharp/SecureMemory/SecureMemory.Tests/ProtectedMemoryImpl/Libc/LibcProtectedMemoryAllocatorTest.cs
@@ -134,7 +134,7 @@ namespace GoDaddy.Asherah.SecureMemory.Tests.ProtectedMemoryImpl.Libc
         {
             Skip.If(libc == null);
 
-            Debug.WriteLine("LibcProtectedMemoryAllocatorTest.TestCheckPointerWithRegularPointerShouldSucceed");
+            Debug.WriteLine("LibcProtectedMemoryAllocatorTest.TestFreeWithInvalidLengthShouldFail");
 
             IntPtr fakePtr = IntPtr.Add(IntPtr.Zero, 1);
             Assert.Throws<LibcOperationFailedException>(() => libcProtectedMemoryAllocator.Free(fakePtr, 0));
--- a/csharp/SecureMemory/SecureMemory.Tests/ProtectedMemoryImpl/ProtectedMemorySecretTest.cs
+++ b/csharp/SecureMemory/SecureMemory.Tests/ProtectedMemoryImpl/ProtectedMemorySecretTest.cs
@@ -148,6 +148,60 @@ namespace GoDaddy.Asherah.SecureMemory.Tests.ProtectedMemoryImpl
 
         [Theory]
         [ClassData(typeof(AllocatorGenerator))]
+        private void TestWithSecretIntPtrSuccess(IProtectedMemoryAllocator protectedMemoryAllocator)
+        {
+            Debug.WriteLine("TestWithSecretUtf8CharsSuccess");
+            char[] secretChars = { 'a', 'b' };
+            using (ProtectedMemorySecret secret =
+                ProtectedMemorySecret.FromCharArray(secretChars, protectedMemoryAllocator, configuration))
+            {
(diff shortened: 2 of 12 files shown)


jgowdy:
The code coverage is as high as I can rationally get on this.  If you look at the lines it's complaining about they're not easily reproduced error checks.  Overall the PR improves code coverage, it's just complaining about percentage of diff.

csharp/SecureMemory/SecureMemory.Tests/ProtectedMemoryImpl/Libc/LibcProtectedMemoryAllocatorTest.cs
```diff
@@ -129,6 +129,17 @@ private void TestCheckPointerWithRegularPointerShouldSucceed()
             }
         }
 
+        [SkippableFact]
+        private void TestFreeWithInvalidLengthShouldFail()
+        {
+            Skip.If(libc == null);
+
+            Debug.WriteLine("LibcProtectedMemoryAllocatorTest.TestCheckPointerWithRegularPointerShouldSucceed");
```
aka-bo, line 137: copypasta
```suggestion
            Debug.WriteLine("LibcProtectedMemoryAllocatorTest.TestFreeWithInvalidLengthShouldFail");
```

csharp/SecureMemory/SecureMemory.Tests/ProtectedMemoryImpl/ProtectedMemorySecretTest.cs
```diff
@@ -146,6 +146,60 @@ private void TestWithSecretUtf8CharsSuccess(IProtectedMemoryAllocator protectedM
             }
         }
 
+        [Theory]
+        [ClassData(typeof(AllocatorGenerator))]
+        private void TestWithSecretIntPtrSuccess(IProtectedMemoryAllocator protectedMemoryAllocator)
+        {
+            Debug.WriteLine("TestWithSecretUtf8CharsSuccess");
```
aka-bo, line 153:
```suggestion
            Debug.WriteLine("TestWithSecretIntPtrSuccess");
```

csharp/SecureMemory/SecureMemory.Tests/ProtectedMemoryImpl/ProtectedMemorySecretTest.cs
```diff
@@ -146,6 +146,60 @@ private void TestWithSecretUtf8CharsSuccess(IProtectedMemoryAllocator protectedM
...
+                    return true;
+                });
+            });
+        }
+
+        [Theory]
+        [ClassData(typeof(AllocatorGenerator))]
+        private void TestWithSecretIntPtrActionSuccess(IProtectedMemoryAllocator protectedMemoryAllocator)
+        {
+            Debug.WriteLine("TestWithSecretUtf8CharsSuccess");
```
aka-bo, line 190:
```suggestion
            Debug.WriteLine("TestWithSecretIntPtrActionSuccess");
```

csharp/SecureMemory/SecureMemory.Tests/SecretTest.cs
```diff
@@ -58,5 +58,25 @@ private void TestWithSecretUtf8CharsActionOfChar()
             secretMock.Object.WithSecretUtf8Chars(actionWithSecret);
             Debug.WriteLine("TestWithSecretUtf8CharsActionOfChar: Finish\n");
         }
+
+        [Fact]
+        private void TestWithSecretIntPtrAction()
+        {
+            Debug.WriteLine("\nTestWithSecretIntPtrActionOfChar: Start");
```
aka-bo, line 65:
```suggestion
            Debug.WriteLine("\nTestWithSecretIntPtrAction: Start");
```

csharp/SecureMemory/SecureMemory/ProtectedMemoryImpl/Libc/LibcProtectedMemoryAllocatorLP64.cs
```diff
@@ -50,13 +48,6 @@ public virtual void SetReadWriteAccess(IntPtr pointer, ulong length)
         // ************************************
         public virtual IntPtr Alloc(ulong length)
         {
-            libc.getrlimit(GetMemLockLimit(), out var rlim);
-            if (rlim.rlim_max != rlimit.UNLIMITED && rlim.rlim_max < length)
-            {
-                throw new MemoryLimitException(
-                    $"Requested MemLock length exceeds resource limit max of {rlim.rlim_max}");
-            }
-
```
aka-bo, line 59: Why was this check removed?

aka-bo (changes requested):
Locking looks good, just a few copy/paste debug logs to clean up

csharp/SecureMemory/SecureMemory/ProtectedMemoryImpl/Libc/LibcProtectedMemoryAllocatorLP64.cs
jgowdy: This check is comparing the global memory limit against the individual allocation size.  It would make sense if we tracked the number of bytes we allocated/deallocated and added that value to length and THEN compared it to the global mlock limit, but otherwise it only catches one unlikely case, where any *single* allocation exceeds the entire mlock limit.

aka-bo: I agree that this check is of limited value, but in the unlikely case of the single allocation, wouldn't this message be more valuable than the dreaded `Libc call mlock failed with result -1 12` that would occur otherwise?

jgowdy: I've replaced the check with tracking of actual memory locked and comparing that value to the limit.

aka-bo (approved):
LGTM

PR #326, opened 2020-10-17
Secure memory refactor

jgowdy:
This is the long awaited refactor of the poor abstractions of SecureMemory


csharp/SecureMemory/PlatformNative/SystemInterface.cs
```diff
@@ -0,0 +1,98 @@
...
+                    {
+                        systemInterface = new MacOSSystemInterfaceImpl();
+                    }
+                    else if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                    {
+                        systemInterface = new WindowsSystemInterfaceImpl();
+                    }
+                    else
+                    {
+                        throw new PlatformNotSupportedException("Unknown platform");
```
rleverence-gd, line 62: Put the platform in the exception message? It's available via `RuntimeInformation.OSDescription`

csharp/SecureMemory/SecureMemory/ProtectedMemoryImpl/OpenSSL/OpenSSL11ProtectedMemoryAllocatorLP64.cs
```diff
@@ -0,0 +1,193 @@
...
+            if (disposedValue)
+            {
+                throw new Exception("Called SetReadAccess on disposed LinuxOpenSSL11ProtectedMemoryAllocatorLP64");
+            }
+
+            Check.IntPtr(pointer, "SetReadAccess");
+
+            // Per page-protections aren't possible with the OpenSSL secure heap implementation
+            // Round up allocation size to nearest block size
+            // Debug.WriteLine($"SetReadAccess: Rounding length {length} to nearest blocksize");
```
rleverence-gd, line 109: Commented out code

csharp/SecureMemory/SecureMemory/ProtectedMemoryImpl/OpenSSL/OpenSSL11ProtectedMemoryAllocatorLP64.cs
```diff
@@ -0,0 +1,193 @@
...
+        public void Dispose()
+        {
+            disposedValue = true;
+            ReleaseUnmanagedResources();
+            GC.SuppressFinalize(this);
+        }
+
+        private void ReleaseUnmanagedResources()
+        {
+            // TODO release unmanaged resources here
```
rleverence-gd, line 190: Is this a critical thing to have done in this PR?

csharp/SecureMemory/SecureMemory/ProtectedMemoryImpl/Windows/WindowsProtectedMemoryAllocatorLLP64.cs
```diff
@@ -1,93 +1,114 @@
...
-        public void SetNoAccess(IntPtr pointer, ulong length)
+        public WindowsProtectedMemoryAllocatorLLP64(IConfiguration configuration, SystemInterface systemInterface, IMemoryEncryption memoryEncryption)
         {
-            length = AdjustLength(length);
+            this.systemInterface = systemInterface ?? throw new ArgumentNullException(nameof(systemInterface));
+            this.memoryEncryption = memoryEncryption;
+            encryptedMemoryBlockSize = memoryEncryption.GetEncryptedMemoryBlockSize();
 
-            if (!WindowsInterop.CryptProtectMemory(pointer, (UIntPtr)length, CryptProtectMemoryOptions.SAME_PROCESS))
+            /*
```
rleverence-gd, line 22: Commented out code.

rleverence-gd (changes requested):

jgowdy:
I think there's an issue here where I'm treating the IV field as an actual IV in OpenSSLCryptProtectMemory rather than the fact that in AEAD it's repurposed as a nonce field.  I think this class suffers from the dreaded nonce reuse for the same key issue.

csharp/SecureMemory/PlatformNative/OpenSSL/OpenSSLCryptoWindows.cs
```diff
@@ -0,0 +1,329 @@
...
+        }
+
+        [DllImport(LibraryName, EntryPoint = "EVP_CIPHER_block_size", SetLastError = true)]
+        private static extern int _EVP_CIPHER_block_size(IntPtr e);
+
+        public int EVP_CIPHER_block_size(IntPtr e)
+        {
+            int blockSize = _EVP_CIPHER_block_size(e);
+
+            // BUG: EVP_CIPHER_block_size returns 1
```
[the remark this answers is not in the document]
jgowdy: I've figured this out, and this actually isn't a bug.  With AEAD ciphers like GCM there is no padding which is why EVP_CIPHER_block_size returns 1 byte.  With AEAD, if you encrypt 2 bytes, you get 2 bytes, there's no rounding to the next block size like with CBC.

jgowdy:
Some block modes (like CBC) require the input to be split into blocks and the final block to be padded to the block size using a padding algorithm (e.g. add a special padding character). Other block modes (like CTR, CFB, OFB, CCM, EAX and GCM) do not require padding at all, because they perform XOR between portions of the plaintext and the internal cipher's state at each step.

jgowdy:
I'm changing direction on this

PR #333, opened 2020-12-15
Set DateParseHandling to None

sushantmimani:
- When retrieving data from a JSON blob, Newtonsoft.Json attempts to parse any data that looks like dates internally as a DateTime. As a result of this, when a date time value is pulled as a string, it uses its own default formatting to make the value different from how it originally was.
To avoid this, we are explicitly setting DateParseHandling to None.
- Bump version for release


aka-bo (approved):
LGTM

PR #339, opened 2021-01-27
[Go] store/load persistence model

aka-bo:
This PR adds support for the custom persistence model (store/load methods) to the go SDK.


aka-bo:
New PR incoming shortly

PR #340, opened 2021-01-29
[Go] store/load persistence model, plus Encrypt/Decrypt API update

aka-bo:
This PR includes two significant changes:
1) Adds support for the custom persistence model (store/load methods) to the go SDK.
2) The session `Encrypt` and `Decrypt` method signatures have changed to include a `context.Context` argument and the now redundant `EncryptContext` and `DecryptContext` methods have been removed.
  ***Note that this change is a NOT backwards compatible and existing code will need to be modified when upgrading to the new version.***


sushantmimani (approved):
LGTM 🦅

PR #359, opened 2021-04-20
[C#] Fix handling of very large/unlimited memlock rlimit

### LINKED ISSUE #358: Csharp GoDaddy.Asherah.AppEncryption 2.1 does not run on Fargate container
**Describe the bug**
CSharp GoDaddy.Asherah.AppEncryption Asherah 0.2.1 does not run on Fargate. Fails with error:
`GoDaddy.Asherah.SecureMemory.ProtectedMemoryImpl.MemoryLimitException: Requested MemLock length exceeds resource limit max of 0
`
Asherah 0.1.6 works fine though. So this is a regression bug.

**To Reproduce**
1. Call decrypt using GoDaddy.Asherah.AppEncryption Asherah 0.2.1 from an ECS container fails.

**Expected behavior**
Decrypt call should return the decrypted content.

**Screenshots**

**Additional context**

aka-bo:
The previous implementation suffered from two issues:
- an rlimit of unlimited was managed internally as a resource limit of 0,
  leading to a MemoryLimitException on subsequent Alloc() calls.
- an rlimit > long.MaxValue resulted in an overflow condition and a resource
  limit of less than 0, also leading to a MemoryLimitExeption.

Resolves #358

### DIFF
--- a/csharp/SecureMemory/SecureMemory/ProtectedMemoryImpl/Libc/LibcProtectedMemoryAllocatorLP64.cs
+++ b/csharp/SecureMemory/SecureMemory/ProtectedMemoryImpl/Libc/LibcProtectedMemoryAllocatorLP64.cs
@@ -18,15 +18,14 @@ namespace GoDaddy.Asherah.SecureMemory.ProtectedMemoryImpl.Libc
         protected LibcProtectedMemoryAllocatorLP64(LibcLP64 libc)
         {
             this.libc = libc ?? throw new ArgumentNullException(nameof(libc));
-
-            libc.getrlimit(GetMemLockLimit(), out var rlim);
-            if (rlim.rlim_max == rlimit.UNLIMITED)
+            var rlim = GetMemlockResourceLimit();
+            if (rlim == rlimit.UNLIMITED || rlim > long.MaxValue)
             {
-                resourceLimit = 0;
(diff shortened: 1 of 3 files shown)


csharp/SecureMemory/SecureMemory/ProtectedMemoryImpl/Libc/LibcProtectedMemoryAllocatorLP64.cs
```diff
@@ -18,15 +18,14 @@ internal abstract class LibcProtectedMemoryAllocatorLP64 : IProtectedMemoryAlloc
         protected LibcProtectedMemoryAllocatorLP64(LibcLP64 libc)
         {
             this.libc = libc ?? throw new ArgumentNullException(nameof(libc));
-
-            libc.getrlimit(GetMemLockLimit(), out var rlim);
-            if (rlim.rlim_max == rlimit.UNLIMITED)
+            var rlim = GetMemlockResourceLimit();
+            if (rlim == rlimit.UNLIMITED || rlim > long.MaxValue)
```
jgowdy, line 22: This is a signed / unsigned comparison for the second value.  I think it's okay and you have a test that covers that path, but I'm making a note of it.

jgowdy (approved):

PR #363, opened 2021-04-26
Fix #362: Records created in another region aren't able to be decry…

### LINKED ISSUE #362: Go + DynamoDB Global Tables- records created in one region fail to decrypt in other regions
Currently, the suffixedPartition does a check (IsValidIntermediateKeyID) to determine if the IK ID is valid. However this fails if trying to read a record encrypted in another region, even though the keys are available in the data store.

**To Reproduce**
1. Create a record in (example) us-west-2. IK gets an ID like `_IK_partid_service_product_us-west-2`
2. Read the same record in (example) us-east-1. The record is read, and the existing IK is compared to:
The current region's partition: `_IK_partid_service_product_us-east-1`. Does not match.
The value without the suffix `_IK_partid_service_product`. Does not match.
Since neither matches, an error is thrown.

**Expected behavior**
Since the IK is present in the datastore, I would expect we would be able to retrieve it and decrypt the record, even though it was created in another region.

levisaya:
…pted with the go appencrpytion implementation

### DIFF
--- a/go/appencryption/partition.go
+++ b/go/appencryption/partition.go
@@ -2,7 +2,7 @@ package appencryption
 
 import (
 	"fmt"
-    "strings"
+	"strings"
 )
 
 func newPartition(partition, service, product string) defaultPartition {
--- a/go/appencryption/partition.go
+++ b/go/appencryption/partition.go
@@ -2,6 +2,7 @@ package appencryption
 
 import (
 	"fmt"
+    "strings"
 )
 
 func newPartition(partition, service, product string) defaultPartition {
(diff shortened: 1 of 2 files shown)


aka-bo (changes requested):
Thanks for the fix @levisaya.
The change looks good, just need to clean up the formatting a bit (whitespace). Running `gofmt` on the modified files should do the trick.

aka-bo (approved):
LGTM

aka-bo:
Resolves #362

PR #377, opened 2021-06-15
Add caching step and use custom token in release jobs

sushantmimani:
Add a caching step in all release jobs so that cached dependencies are downloaded. This change also changes the default github token used in the release jobs. The reason for this is that the default token does not trigger dependent workflows.
https://github.community/t/github-actions-workflow-not-triggering-with-tag-push/17053


tarkatronic (approved):

aka-bo (approved):
LGTM

PR #382, opened 2021-07-06
Update C# AE csproj file

sushantmimani:
Currently, the CI for C# AppEncryption fails with the following error.
```
/home/runner/work/asherah/asherah/csharp/AppEncryption/AppEncryption/AppEncryption/obj/Release/netstandard2.1/.NETStandard,Version=v2.1.AssemblyAttributes.cs(4,12): error CS0579: Duplicate 'global::System.Runtime.Versioning.TargetFrameworkAttribute' attribute
```
The proposed solution is derived from this SO article - https://stackoverflow.com/questions/61997928/errorcs0579duplicate-globalsystem-runtime-versioning-targetframeworkattribu

Verified that the solution works here - https://github.com/sushantmimani/asherah/runs/3000940891?check_suite_focus=true
The failure is due to missing secrets in the fork.


tarkatronic (approved):
Seems reasonable to me.

Issue #385, opened 2021-07-20
[Go] Golang: Degraded, unrecoverable performance of DRR decrypt times after RevokeCheckInterval expires.


keperry:
**Describe the bug**
Golang: Degraded, unrecoverable performance of DRR decrypt times after RevokeCheckInterval expires.

We noticed some large increases in DRR decrypt times - 99 percentile of `ael.drr.decrypt` (we were using the metrics from the library w/prometheus and grafana).  It appeared that the app would never recover.  We were trying to upgrade from an older version (not in this repo) to the latest version v0.2.1
See screenshot below:
![DataFabric_-_Encryption_Database_Metrics_-_Grafana_and_mgntapi_–_encryption_go](https://user-images.githubusercontent.com/1618767/126328322-5b59a204-3979-41a7-848e-81fed03f3902.png)

**To Reproduce**
1.  Use default crypto policy (RevokeCheckInterval at 60m).  You can reduce this to reduce test timing.  
2. Startup app using this that uses decrypt
3.  At the RefreshCheckInterval, you will see DRR times skyrocket (10x?).  They seem to keep climbing until the app is unusable.

Note:  In PR, I will have a test and benchmark test to reproduce.

**Expected behavior**
We expect that after the RevokeCheckInterval expires that there is a slight hit in performance when the new system/intermediate keys are loaded, but the app should be usable after this point.

**Screenshots**
See above.

PR coming soon.

tarkatronic:
Fixed in #386

PR #386, opened 2021-07-20
[Go] Fix performance issues with expired cache (decrypt) - issue #385

keperry:
This should fix issue #385 
It appears that when the cache item was expired and then reloaded, it was never written back to the cache.  This caused the item to get loaded every time after RevokeCheckInterval  expired.

There is a small fix for writing back to the cache, and then I also moved the locks around to load/get.   This should allow for better performance on reads with multiple readers.  I also ran "go mod tidy".  I added a benchmark test and normal test so we can test this in the future. 
To run the benchmark, try something like this:
`go test -run="notthis" -bench='^BenchmarkKeyCache_GetOrLoad_MultipleThreadsRead_NeedReloadKey$' -v -cpu=12 -cpuprofile=cpu.out -benchtime=60s 
`
I noticed significant improvements after making the "write" fix.  After making the "write" fix + "lock changes", I found a slight improvement as well.

CPU profiles:

- Original: `BenchmarkKeyCache_GetOrLoad_MultipleThreadsRead_NeedReloadKey-12           12675           5642228 ns/op`

- With Write Fix Only: `BenchmarkKeyCache_GetOrLoad_MultipleThreadsRead_NeedReloadKey-12        98069263              1094 ns/op`

- With Write Fix And Lock Changes: `BenchmarkKeyCache_GetOrLoad_MultipleThreadsRead_NeedReloadKey-12        100000000              803.6 ns/op`

I have attached the [cpuprofiles.zip](https://github.com/godaddy/asherah/files/6849304/cpuprofiles.zip)
 for all three scenarios.

1. Original
2. Write fix only
3. Write fix with lock changes

Here we are comparing the original with the write fix only:
`go tool pprof -nodefraction=0 -http=:8086 -normalize -diff_base cpu_original.out cpu_write_fix.out 
`cpu_diff_write_fix

Here is comparing write fix with write fix and lock changes : 
`go tool pprof -nodefraction=0 -http=:8086 -normalize -diff_base cpu_write_fix.out cpu_write_fix_with_lock_changes.out 
`image

We ran a small canary test in prod.  With the RevokeCheckInterval set to 5 minutes, we saw much better performance with this fix.
Datafabric_DRR_metrics_after_fix


go/appencryption/cache.go
```diff
@@ -96,9 +96,6 @@ func isReloadRequired(entry cacheEntry, checkInterval time.Duration) bool {
 // is not present in the cache it will retrieve the key using the provided keyLoader
 // and store the key if an error is not returned.
 func (c *keyCache) GetOrLoad(id KeyMeta, loader keyLoader) (*internal.CryptoKey, error) {
-	c.rw.Lock()
-	defer c.rw.Unlock()
-
```
aka-bo, line 101: While locking at this level does incur a performance cost (as you've found) it is necessary. Moving the lock to the get/lock methods introduces a deadlock scenario.

go/appencryption/cache.go
```diff
@@ -112,6 +109,9 @@ func (c *keyCache) GetOrLoad(id KeyMeta, loader keyLoader) (*internal.CryptoKey,
 // The second return value indicates the successful retrieval of a
 // fresh key.
 func (c *keyCache) get(id KeyMeta) (*internal.CryptoKey, bool) {
+	c.rw.RLock()
+	defer c.rw.RUnlock()
+
```
aka-bo, line 114: as noted above, this needs to be reverted.

go/appencryption/cache.go
```diff
@@ -127,6 +127,9 @@ func (c *keyCache) get(id KeyMeta) (*internal.CryptoKey, bool) {
 // load maintains the latest entry for each distinct ID which can be accessed using
 // id.Created == 0.
 func (c *keyCache) load(id KeyMeta, loader keyLoader) (*internal.CryptoKey, error) {
+	c.rw.Lock()
+	defer c.rw.Unlock()
+
```
aka-bo, line 132: revert

aka-bo (changes requested):
Hi @keperry and thank you for the PR and excellent write-up describing the problem!

The underlying fix looks good but as noted in the comments the locking changes need to be reverted as they introduce a race (deadlock).

keperry:
Note:  Modified locks to use RLock and Lock on GetOrLoad().  This should get us both quick reads and performant writes.  I also added an early exit for the heavy "Lock()" - that should prevent several queued up RW locks from all performing loads.   Here are the profiles for both scenarios:  
[cpuprofiles_take2.zip](https://github.com/godaddy/asherah/files/6851021/cpuprofiles_take2.zip)


Here are some results from my machine - the performance looks a lot better with the lock changes:
- Write Fix only: 
`BenchmarkKeyCache_GetOrLoad_MultipleThreadsRead_NeedReloadKey-12        129446646              585.7 ns/op`
- Write Fix with Lock Improvements:
`BenchmarkKeyCache_GetOrLoad_MultipleThreadsRead_NeedReloadKey-12        551315848              177.4 ns/op
`
image

aka-bo (approved):
LGTM

aka-bo:
Merging... I'll take a closer look at the unrelated build issue as part of the release process.

Thanks again for the PR!

PR #390, opened 2021-09-01
Option to set Aurora replica read consistency

jeffchannell:
This PR adds an option to the Golang sidecar server to optionally set the `aurora_replica_read_consistency` parameter to allow global write forwarding to function correctly:

https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/aurora-global-database-write-forwarding.html

Without this patch, a sidecar connected to a secondary cluster with write forwarding enabled will fail because it cannot write to the database. With this patch, the database session is primed with the value given in the new option and writes function correctly.

### DIFF
--- a/server/go/pkg/server/database_test.go
+++ b/server/go/pkg/server/database_test.go
@@ -1,19 +1,60 @@
 package server
 
 import (
+	"context"
 	"database/sql"
 	"database/sql/driver"
+	"fmt"
 	"os"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 )
 
+var (
+	readConsistencyValue string
+)
+
+type FakeResult struct {}
+
+func (r *FakeResult) LastInsertId() (int64, error) {
+	return 0, nil
+}
+
+func (r *FakeResult) RowsAffected() (int64, error) {
+	return 0, nil
+}
+
+type FakeConn struct {
+}
+
+func (c *FakeConn) Begin() (driver.Tx, error) {
+	return nil, nil
+}
+
+func (c *FakeConn) Close() error {
+	return nil
+}
+
+func (c *FakeConn) ExecContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Result, error) {
+	result := &FakeResult{}
+	switch query {
+	case ReplicaReadConsistencyQuery:
+		readConsistencyValue = fmt.Sprintf("%v", args[0].Value)
+	}
+	return result, nil
+}
+
+func (c *FakeConn) Prepare(query string) (driver.Stmt, error) {
+	return nil, nil
+}
+
 type FakeDriver struct {
 }
 
 func (f *FakeDriver) Open(name string) (driver.Conn, error) {
-	return nil, nil
+	conn := &FakeConn{}
+	return conn, nil
 }
 
 func TestMain(m *testing.M) {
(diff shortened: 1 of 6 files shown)


aka-bo (changes requested):
Hi @jeffchannell, thanks for the PR!

Additional reviewer feedback aside, I just have one housekeeping request for you that'll need to be taken care of prior to merge. At present, this PR includes merge commits that don't belong upstream which will need to be removed. The easiest way to pull this off can vary but a rebase and force push usually does the trick.

Here's a Stack Overflow Q&A outlining a few options: https://stackoverflow.com/questions/33502205/how-to-delete-merge-commits

server/go/pkg/server/database_test.go
```diff
@@ -32,3 +75,29 @@ func TestNewMysql(t *testing.T) {
...
 	assert.NotNil(t, dbconnection)
 }
+
+func TestSetRdbmsReplicaReadConsistencyValue(t *testing.T) {
+	dbconnection = nil
+
+	var err error
+	db, err := newMysql("root:secret@(localhost:%s)/mysql")
+	assert.Nil(t, err)
+	assert.NotNil(t, db)
```
tarkatronic, line 85: Two questions here:

`dbconnection` seems to be unused except above in `TestNewMysql`. TBH I can't even tell where this variable is coming from. But here in this test, you're using `db`. So why the reference to `dbconnection`?

Second, I don't think these first two assertions are necessary since they are just duplicating `TestNewMysql` above, verbatim.

server/go/pkg/server/database_test.go
```diff
@@ -32,3 +75,29 @@ func TestNewMysql(t *testing.T) {
...
+	assert.Nil(t, err)
+	assert.Equal(t, ReplicaReadConsistencyValueEventual, readConsistencyValue)
+	// global
+	err = setRdbmsReplicaReadConsistencyValue(ReplicaReadConsistencyValueGlobal)
+	assert.Nil(t, err)
+	assert.Equal(t, ReplicaReadConsistencyValueGlobal, readConsistencyValue)
+	// session
+	err = setRdbmsReplicaReadConsistencyValue(ReplicaReadConsistencyValueSession)
+	assert.Nil(t, err)
+	assert.Equal(t, ReplicaReadConsistencyValueSession, readConsistencyValue)
```
tarkatronic, line 102: Given that you are asserting 4 entirely different scenarios here, can this be split into 4 tests? I recognize they would all have the same setup, but the different cases should be allowed to fail separately.

tarkatronic (commented):
Thanks for the contribution @jeffchannell! I just had a couple of small nits and questions on the test suite. Aside from that, I'm excited to get this in!

server/go/pkg/server/database_test.go
tarkatronic: Ah ignore the first part on `dbconnection` -- I missed this at the package level initially. 😄

jeffchannell: I can remove the additional assertions, and only really had them in there to ensure the correct state for the test.

server/go/pkg/server/database_test.go
jeffchannell: I've split the test and cleaned up a wee bit along the way.

tarkatronic (approved):
🎉

aka-bo (approved):
LGTM

aka-bo:
Thanks again for the contribution!

PR #392, opened 2021-09-15
Add an implementation of SecureMemory that doesn't use mlock

sushantmimani:
Add a new implementation of SecureMemory which:

- Doesn't use mlock
- Ensures that values stored will not show up in core dumps
- Ensures that values stored will be securely / explicitly zeroed out when no longer in use

Currently supports Linux & MacOS


lyricnz:
How does this relate to the significant work that was done in https://github.com/godaddy/asherah/pull/300 ?

ash-manos:
If there is any way someone can review and merge this ASAP we would be grateful.

@sushantmimani do you know whether any of your colleagues can review it? We have an urgent Production outage because of the mlock issues occurring in a lambda runtime. We've forked and tested this implementation of yours and it's resolved the issue - but I'm not sure we have the in-depth knowledge of this library to review your changes with full confidence.

aka-bo:
> If there is any way someone can review and merge this ASAP we would be grateful.

I've been working with @sushantmimani offline and with any luck we'll get this wrapped up later today.

csharp/SecureMemory/SecureMemory/SecureMemorySecretFactory.cs
```diff
@@ -180,11 +183,12 @@ private static IProtectedMemoryAllocator DetectViaRuntimeInformation(IConfigurat
...
         }
 
-        private static IProtectedMemoryAllocator ConfigureForLinux64(IConfiguration configuration)
+        private static ISecureMemoryAllocator ConfigureForLinux64(IConfiguration configuration)
         {
             if (configuration != null)
             {
-                var secureHeapEngine = configuration["secureHeapEngine"];
+                string secureHeapEngine = configuration["secureHeapEngine"];
+                bool.TryParse(configuration["mlock"], out bool mLock);
```
aka-bo, line 191: Need to ensure the use of `mlock` is the default behavior of this library, but this seems to require not only the use of a configuration (which should be optional) but the "mlock" config setting must be set to "true" to enable the `mlock` protection.

Once quick fix would be to use "enabled" and "disabled" (defaulting of "enabled" of course) in place of "true"/"false"

csharp/SecureMemory/SecureMemory.Tests/SecureMemorySecretFactoryTest.cs
```diff
@@ -0,0 +1,108 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using Microsoft.Extensions.Configuration;
+using Xunit;
+
+namespace GoDaddy.Asherah.SecureMemory.Tests
+{
+    [Collection("Logger Fixture collection")]
+    public class SecureMemorySecretFactoryTest
```
aka-bo, line 10: Needs tests for the new `mlock` configuration setting

aka-bo (changes requested):
Looking good overall, just a couple of concerns regarding the new config setting (see comments)

aka-bo (approved):
LGTM

csharp/SecureMemory/SecureMemory/SecureMemoryImpl/Linux/LinuxSecureMemoryAllocatorLP64.cs
```diff
@@ -0,0 +1,104 @@
...
+            this.libc = libc;
+        }
+
+        public override void Dispose()
+        {
+        }
+
+        internal override int GetRlimitCoreResource()
+        {
+            return (int)RlimitResource.RLIMIT_CORE;
```
jgowdy, line 45: If I understand correctly, the protected memory allocator version locks, but the secure memory allocator does not.  If that's the case, we can/should probably drop the rlimit and locking related aspects out of the secure memory allocator right?

jgowdy (approved):
I don't see anything blocking on this PR.  👍🏻

csharp/SecureMemory/SecureMemory/SecureMemoryImpl/Linux/LinuxSecureMemoryAllocatorLP64.cs
aka-bo: Right, but `RLIMIT_CORE` is used to restrict the size of core dump files (which we use to disable core dumps), so this one doesn't really fall into the "rlimit and locking" category.

csharp/SecureMemory/SecureMemory/SecureMemoryImpl/Libc/LibcSecureMemoryAllocatorLP64.cs
```diff
@@ -0,0 +1,59 @@
...
+            try
+            {
+                // Wipe the protected memory (assumes memory was made writeable)
+                ZeroMemory(pointer, length);
+            }
+            finally
+            {
+                try
+                {
+                    Interlocked.Add(ref memoryLocked, 0 - (long)length);
```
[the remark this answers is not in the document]
sushantmimani: Leftover from copy-pasta. Deleted now.

lyricnz:
Great work!

PR #393, opened 2021-09-15
Test fixes for multiple platforms

aka-bo:
In addition to the test fixes for multiple platforms (mainly macOS), `ProtectedMemorySecretFactory` will now throw a `PlatformNotSupportedException` when configured with `secureHeapEngine=openssl11` on macOS, instead of silently falling back to `mmap`.


sushantmimani (approved):
LGTM 🦅

PR #400, opened 2022-01-06
[Go] Fix for intermittent encryption errors

bdittmer-godaddy:
We have seen intermittent errors encrypting a payload.  The error message is:
error decrypting data: cipher: message authentication failed

In this case, it appears that an IK rotates and uses an SK that is about to expire.  Next request, the SK expires so it creates a new one but cannot create a new IK for an interval up to the CreateDatePrecision in the specified policy as that would create a duplicate.


aka-bo (approved):
LGTM

PR #409, opened 2022-03-17
Do not assume localhost for go cross-language test

jgowdy:


jgowdy:
When leveraging a container to run the tests in a downstream repo, I need to be able to honor the MYSQL_HOSTNAME variable rather than assuming we're running on the host and can access via localhost.

Thanks!

aka-bo (approved):
LGTM

PR #505, opened 2022-10-04
Explicitly update to System.Text 6.x packages

jgowdy:


jgowdy:
Let's see if this fixes the vulnerability in App.Metrics

aka-bo (approved):

PR #871, opened 2023-05-02
Workflow Caching and Test Improvements

aka-bo:
## Description

CI jobs have been updated to use caching provided by the `setup-{java,go,node}` actions. Note that `actions/setup-dotnet` does not provide this feature, so dotnet packages continue to be cached via `actions/cache`.

Also included in this PR are a number of improvements specific to Go:
- [CI] install `gotestsum@latest` and use `--format testname` when running tests
- [CI] do not run integration test packages in parallel
- [integrationtest] use require to prevent noise, use suite assertions
- [integrationtest] logging improvements

And, last but not least, suppressed Maven transfer progress logging accoss the board (`--no-transfer-progress`).


build/go/test.sh
```diff
@@ -2,4 +2,4 @@
 #Using https://github.com/gotestyourself/gotestsum for reference
 
 # Looks like CGO has to be enabled for -race if we're using go modules
-CGO_ENABLED=1 gotestsum --junitfile junit_results.xml -- -race -coverprofile coverage.out -v ./...
+CGO_ENABLED=1 gotestsum -f testname --junitfile junit_results.xml -- -race -coverprofile coverage.out -v ./...
```
sushantmimani, line 5:
```suggestion
gotestsum -f testname --junitfile junit_results.xml -- -race -coverprofile coverage.out -v ./...
```

sushantmimani (approved):
One minor update. Looks good otherwise! 🚀

build/go/test.sh
aka-bo: 👍

PR #894, opened 2023-05-25
Upgrade AWS SDK to v2

agabeev-godaddy:
This PR upgrades AWS SDK to latest v2 version used in java wrapper.


aka-bo (changes requested):
Hi @agabeev-godaddy, thanks for the PR! The move to AWS SDK v2 has been on my personal wishlist for ages. Regarding the builder changes, while I like the idea of a configurable client, we need to avoid breaking API changes.

One idea that comes to mind would be to leave the constructor as-is, default to a client initialized by the builder (similar to how it’s done today), and implement a new build step to optionally configure the client, e.g., `DynamoDbMetastoreImpl.newBuilder("us-west-2”).withClient(dynamoDbClient).build()`. That said, I’m open to other suggestions you may have for moving forward while maintaining backwards compatibility.

Additionally, the reference-app and server projects are intended to target published versions only, so these changes will need to be reverted/removed from the PR (same goes for new install script and associated workflow update).

agabeev-godaddy:
Hi @aka-bo, thanks for review.

I've reverted changes in ci/samples/server module and made changes backward compatible.

aka-bo (approved):
Thank you for the contribution!

PR #1441, opened 2025-08-02
Logging go bye bye

### LINKED ISSUE #1440: Eliminate usage of GoDaddy.Asherah.Logging
**Is your feature request related to a problem? Please describe.**
There are more standard ways to integrate Logging into libraries than to continue to maintain and use a wrapper library.  We should not use GoDaddy.Asherah.Logging in our libraries, and if logging is required, the libraries can have their own way to use ILoggerFactory and ILogger following standard Microsoft patterns.

**Describe the solution you'd like**
No more dependency on any of the Asherah libraries on GoDaddy.Asherah.Logging
Plan a retirement party (timeline) for supporting GoDaddy.Asherah.Logging

**Describe alternatives you've considered**
Any use of ILoggerFactory or ILogger can be baked into libraries that need it.
Eventually this also becomes more clear with proper DI support

**Additional context**
Currently logging does not seem to be used many places, so it could just be eliminated.

chief-micco:
To help us get this pull request reviewed and merged quickly, please be sure to include the following items:

* [x] Tests (if applicable)
* [x] Documentation (if applicable)
* [x] A full explanation here in the PR description of the work done

## PR Type
What kind of change does this PR introduce?

* [ ] Bugfix
* [x] Feature
* [ ] Code style update (formatting, local variables)
* [ ] Refactoring (no functional changes, no api changes)
* [ ] Build related changes
* [ ] CI related changes
* [ ] Documentation content changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

Is this change backward compatible with the most recently released version? Does it introduce changes which might change the user experience in any way? Does it alter the API in any way?

* [ ] Yes (backward compatible)
* [x] No (breaking changes)

Note: The only breaking changes were to public constructors of the following:
- SessionBytesImpl
- SessionJsonImpl
- EnvelopeEncryptionBytesImpl
- EnvelopeEncryptionJsonImpl
These are not documented as something that consumers use directly, and seem to only have public constructors to facilitate tests. This can be called out.

Another change needed will be that consumers will need to add the `WithLogging` to some of the various builders
and pass in their own ILogger instead of using the GoDaddy.Asherah.Logging.LogManager.SetLoggerFactory

## Issue Linking
fixes #1440 

## What's new?
- This change removes the dependency on GoDaddy.Asherah.Logging.
- It does this by adding a proper `WithLogging` method to the Builders
- If the new `WithLogging` method is not called, log messages will not be written

Other small items included in the change
- Some test files got their test methods updated to be `public` which follows better c# unit test practices
- New private members now (like _logger) will properly use underscore, which is another C# naming best practice


chief-micco:
Review through all the files I changed. No unexpected changes found.

PR #1741, opened 2026-08-10
Centralize NuGet trusted publishing

chief-micco:
To help us get this pull request reviewed and merged quickly, please be sure to include the following items:

* [ ] Tests (if applicable)
* [ ] Documentation (if applicable)
* [ ] Changelog entry
* [x] A full explanation here in the PR description of the work done

## PR Type
What kind of change does this PR introduce?

* [ ] Bugfix
* [ ] Feature
* [ ] Code style update (formatting, local variables)
* [ ] Refactoring (no functional changes, no api changes)
* [ ] Build related changes
* [x] CI related changes
* [ ] Documentation content changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

Is this change backward compatible with the most recently released version? Does it introduce changes which might change the user experience in any way? Does it alter the API in any way?

* [x] Yes (backward compatible)
* [ ] No (breaking changes)

## What's new?
- Updated nuget publishing to use Trusted Publishing
- Centralized the nuget publshing script to ensure it matches the Trusted Publishing expected script


.github/workflows/csharp-release.yml
aka-bo, line 1: Every other workflow here follows `<language>-[<component>-]<purpose>.yml` with a `[<Language>] <Component> <Purpose>` display name — 20 of 21 files. `csharp-pr.yml`, `go-pr.yml` and `java-pr.yml` are the established component-less form for a workflow that spans a whole language, so `csharp-release.yml` / `[C#] Release` fits without inventing a pattern. (`cross-language-validation.yml` is unprefixed because it genuinely spans every language; NuGet is C#-only.)

Two practical reasons to settle it in this PR rather than later:

- NuGet.org trusted publishing policies bind to the workflow filename, so a later rename means re-editing the policy, with publishing broken until that happens.
- `csharp-pr.yml` filters on `.github/workflows/csharp-*.yml`, so as `nuget-release.yml` this file sits outside every C# path filter and future edits to it get no PR validation.

Also needs the file renamed and the three `uses:` references updated:
`csharp-appencryption-ci.yml:130`, `csharp-logging-ci.yml:70`, `csharp-securememory-ci.yml:113`.

.github/workflows/csharp-release.yml
```diff
@@ -0,0 +1,97 @@
...
+        required: true
+        type: string
+      package-directory:
+        description: Directory containing the C# package project and Directory.Build.props
+        required: true
+        type: string
+      release-name:
+        description: Package family name used for the release tag
+        required: true
+        type: string
```
aka-bo, line 17: Suggested new input to support the Logging artifact path fix below.

```suggestion
        type: string
      artifact-path:
        description: Directory to extract the build artifact into (defaults to package-directory)
        required: false
        type: string
```

.github/workflows/nuget-release.yml
```diff
@@ -0,0 +1,97 @@
...
+        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+        with:
+          path: ~/.nuget/packages
+          key: ${{ runner.os }}-nuget-${{ inputs.release-name }}-${{ hashFiles(format('{0}/**/*.csproj', inputs.package-directory)) }}-v2
+
+      - name: Download artifact
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: ${{ inputs.artifact-name }}
+          path: ${{ github.workspace }}/${{ inputs.package-directory }}
```
aka-bo: `upload-artifact` roots an artifact at the least common ancestor of the matched files, and that differs per package. `csharp-app-encryption` and `csharp-secure-memory` each span several project directories, so they root at the package directory — but `csharp-logging` uploads only `csharp/Logging/Logging/{bin,obj}/**`, so it roots at `csharp/Logging/Logging`. A uniform `package-directory` destination therefore extracts Logging's `bin`/`obj` one level too high, so `dotnet pack` silently repacks from a fresh build instead of the artifact that passed CI, and leaves stray `bin`/`obj` trees at the solution root. The previous `csharp-logging-release.yml` used `csharp/Logging/Logging` for this reason.

```suggestion
          path: ${{ github.workspace }}/${{ inputs.artifact-path || inputs.package-directory }}
```

.github/workflows/csharp-release.yml
```diff
@@ -0,0 +1,97 @@
...
+
+permissions:
+  contents: read
```
aka-bo, line 24: The three deleted workflows all set `defaults.run.shell: bash`, and the publish step uses bash-only process substitution (`done < <(find …)`). It works today because bash is the default shell on `ubuntu-latest`, but the explicit default is cheap insurance on a step that pushes packages.

```suggestion
permissions:
  contents: read

defaults:
  run:
    shell: bash
```

.github/workflows/nuget-release.yml
```diff
@@ -0,0 +1,97 @@
...
+
+      - name: Initialize environment variables
+        run: |
+          echo "BASE_VERSION=$(grep -o '<Version>.*<.*>' Directory.Build.props | sed 's/<Version>\(.*\)<.*>/\1/')" >> "$GITHUB_ENV"
+          echo "VERSION_SUFFIX=$(echo "${BASE_VERSION}" | cut -f2 -d'-')" >> "$GITHUB_ENV"
+          echo "BRANCH=${GITHUB_REF#refs/heads/}" >> "$GITHUB_ENV"
```
aka-bo: `$GITHUB_ENV` is applied after the step finishes, so `${BASE_VERSION}` is still unset on the following line — `VERSION_SUFFIX` always ends up empty, and the `!= 'alpha'` guard on the login and publish steps never fires. A `0.11.0-alpha` version on a `release-*` branch would publish to nuget.org. This is pre-existing rather than a regression (the deleted workflows carried the same two lines, and `release_prod.sh` recomputed the version itself), but this is a convenient place to fix it.

The fix is the plain shell assignment on the first line: `BASE_VERSION` becomes a live variable in this shell, so the expansion on the next line resolves, and its value is then written to the env file explicitly. The current code only appends text to the env file, which the runner does not apply until the step exits.

```suggestion
        run: |
          BASE_VERSION=$(grep -o '<Version>.*<.*>' Directory.Build.props | sed 's/<Version>\(.*\)<.*>/\1/')
          {
            echo "BASE_VERSION=${BASE_VERSION}"
            echo "VERSION_SUFFIX=$(cut -f2 -d'-' <<<"${BASE_VERSION}")"
            echo "BRANCH=${GITHUB_REF#refs/heads/}"
          } >> "$GITHUB_ENV"
```

.github/workflows/nuget-release.yml
```diff
@@ -0,0 +1,97 @@
...
+        working-directory: ${{ inputs.package-directory }}
+
+      - name: NuGet login
+        id: nuget-login
+        if: startsWith(env.BRANCH, 'release-') && env.VERSION_SUFFIX != 'alpha'
+        uses: NuGet/login@ebc737b6fc418a6ca0073cf116ec8dc156d8b81e # v1.0.0
```
aka-bo: `ebc737b6…` is the annotated tag object for `refs/tags/v1`, not a commit (`gh api repos/NuGet/login/git/ref/tags/v1` returns `"type": "tag"`), and `NuGet/login` has no `v1.0.0` tag at all — only `v1`, `v1.1.0` and `v1.2.0`. It peels to `8d196754…` = v1.2.0, so it does resolve today, but it is effectively pinned to a moving major-version tag: if `v1` is re-pointed, this object becomes unreachable and the release breaks on an unresolvable action. Dependabot also can't track a pin in this form.

```suggestion
        uses: NuGet/login@8d196754b4036150537f80ac539e15c2f1028841 # v1.2.0
```

.github/workflows/csharp-logging-ci.yml
```diff
@@ -63,8 +63,12 @@ jobs:
...
       contents: write
-    name: Release
-    uses: ./.github/workflows/csharp-logging-release.yml
-    secrets: inherit
+      id-token: write
+    uses: ./.github/workflows/nuget-release.yml
+    with:
+      artifact-name: csharp-logging
+      package-directory: csharp/Logging
+      release-name: Logging
```
aka-bo, line 74: Pairs with the `artifact-path` input above — Logging's artifact is rooted one level deeper than the other two.

```suggestion
      release-name: Logging
      artifact-path: csharp/Logging/Logging
```

aka-bo (changes requested):
Nice cleanup — three near-identical workflows and four scripts down to one reusable workflow, with API-key publishing replaced by OIDC. Scoping `git push origin "$TAG"` and the `PARENT_COMMIT` fallback are good catches. A few things before merge, details inline:

1. `nuget-release.yml:66` — the `NuGet/login` pin is a tag object rather than a commit.
2. `nuget-release.yml:54` — the uniform artifact path is wrong for Logging.
3. Filename/display name depart from the naming convention, and the filename is what NuGet.org's trusted publishing policy binds to, so it is much cheaper to settle now. The NuGet.org trusted publishing policy will need to be updated for the new workflow filename, assuming it has already been created.

Two optional cleanups also noted inline.

aka-bo (approved):
🚀

